PR #22, opened 2024-05-31
data model changes to support step 3 and 4 of registrations

turb0c0w:
*Issue:*
https://github.com/bcgov/entity/issues/21384
https://github.com/bcgov/entity/issues/21380

*Description of changes:*
- separation of contacts from users in the data model
- support for steps 3 + 4 of registration form
- preliminary gcp bucket support for file upload / delete 


strr-api/src/strr_api/services/gcp_storage_service.py
```diff
@@ -0,0 +1,88 @@
...
+    def registration_documents_bucket(cls):
+        """Get gcp bucket for saving or deleting registration documents."""
+
+        # project_id = current_app.config.get("GCP_CS_PROJECT_ID")
+        scope = current_app.config.get("GCP_CS_SA_SCOPE")
+        bucket_id = current_app.config.get("GCP_CS_BUCKET_ID")
+        # auth_key = current_app.config.get("GCP_AUTH_KEY")
+
+        # TODO: fix service account setup
+        service_account_info = {
```
turb0c0w, line 55: @pwei1018 using ppr-api as an example here, it looks like we need a few more environment variables for the gcp bucket connection:  https://github.com/bcgov/ppr/blob/main/ppr-api/src/ppr_api/callback/auth/token_service.py#L48

PR #58, opened 2024-07-17
Bugfix: Adding redirect on account creation with no documents

samIndustrio:
*Issue:*https://github.com/bcgov/entity/issues/

*Description of changes:*

Adding redirect logic to allow redirect to pay after account creation when:
 - no files uploaded
 - file upload fails
 - all files uploaded

Fixing ToS acceptance loop issue

Fixing logic for prepopulating information after account select


dimak1 (approved):
LG!

PR #75, opened 2024-07-29
Backend: Add startDate & expiryDate to Registration

shaangill025:
*Issue:* https://github.com/bcgov/entity/issues/22422

*Description of changes:*
Added startDate and endDate to Registration model, automatically set these dates when registration status is changed to `APPROVED` or `PROVISIONAL`, and other changes. 


dimak1 (approved):
Looks good to me. The only thing I would update is the terminology, instead of `endDate` -> `expiryDate` to make it more clear.

PR #107, opened 2024-08-20
Fix: Swagger for search applications

kris-daxiom:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*


dimak1 (approved):
The change looks good, however, I tried to search with `text` in query param which didn't work - it did not return the searched results.

PR #149, opened 2024-09-22
Add Application Number for Search and Migration Fix

shaangill025:
*Issue:*

- https://github.com/bcgov/entity/issues/23279
- https://github.com/bcgov/entity/issues/23364

*Description of changes:*
- [x] `applicationNumber` added to `application_json` for search
- [x] Migrations `strr-api/migrations/versions/20240908_0710_7a9f3c2e8b5d_.py` and `strr-api/migrations/versions/20240916_1420_8b2e4d1f9c3a_.py` removed
- [x] Migration `strr-api/migrations/versions/20240908_1010_7a9f3c2e8b4e_.py` added
- [x] Change to 14 character long random digits only `application_number`
- [x] nanoid collision logic


shaangill025:
> Even though the chance of application number collision is less, if it happens to generate a duplicate id, will the application number be null?

This should be addressed now.

PR #157, opened 2024-10-01
Changes to support platforms and strata hotels

kris-daxiom:
*Issue:*

- bcgov/entity/issues/23576

*Description of changes:*

- Model changes to support platforms and strata hotels
- Removed the eligibility table and moved the data to rental property
- Changed the documents model foreign key to registration id
- Renamed RentalPlatforms model to PropertyListings
- Updated the services, resources etc. to populate data to the latest model
- Refactored list registrations endpoint
- Updated unit tests


thorwolpert (approved):
Thanks for keeping the model and terms clean!

PR #158, opened 2024-10-01
Add code owner for strr-api directory

avni-work:
*Issue:*

- bcgov/entity/issues/

*Description of changes: adding Lekshmi as a code owner*


rstens (approved):
Do we want to add other codeowners as well?

avni-work:
> Do we want to add other codeowners as well?

i think so, but we can test drive with this and see from here who wants ownership of what

rstens:
> > Do we want to add other codeowners as well?
> 
> i think so, but we can test drive with this and see from here who wants ownership of what

Sounds good.

PR #192, opened 2024-10-17
Platform review step

deetz99:
*Description of changes:*

- done main layout for review step, will add more comments to specific files


strr-pm-web/app/components/form/platform/Review/EditCardHeader.vue
deetz99, line 1: will edit ConnectPageSection with an actions prop to avoid needing to do this

strr-pm-web/app/components/connect/InfoBox.vue
deetz99, line 1: had to add this because sometimes the content would be horizontally aligned

strr-pm-web/app/pages/platform/application.vue
```diff
@@ -125,6 +125,18 @@ const setPreviousStep = () => {
   }
 }
 
+// something like this? need to discuss options
+// how can we set loading state on the submit and pay button?
+// i wonder if tracking the steps in a store would be useful
```
kialj876, line 130: hmm good idea that makes it more intuitive I think. Otherwise we'll need to provide 'loading' as an option in the button config and then we're calling a util function to edit the meta data to toggle it which feels wrong

strr-pm-web/app/pages/platform/application.vue
```diff
@@ -125,6 +125,18 @@ const setPreviousStep = () => {
...
+const handlePlatformSubmit = () => {
+  // validate each step
+  // if invalid step, set prop on Review component to highlight invalid fields/section ?
+  // show alert with some error text ?
```
kialj876, line 134: yeah something like this. I think the worst case would just be scrolling up to the stepper as it has the unfilled steps marked

kialj876 (approved):
Nice work 👍

PR #202, opened 2024-10-21
platform dashboard

kialj876:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*
- Renaming of connect form components is an ask from the connect repo PR (let me know if this is in line with what you were thinking)
- dashboard work is in progress (several things I still need to do), but I was hoping you'd have a look at the way I'm doing the 'DetailsHeader' in the layout and setting the values via a connect store. Was thinking we would update the button control / breadcrumbs to do it this way as well instead of adding custom page meta values, but have a look and see if you like it
- new page path: .../platform/registration/:regId
  - example with real dev platform registration: https://strr-platform-dev--pr-202-kinxn4mf.web.app/en-CA/platform/registration/131


bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-202-kinxn4mf.web.app

deetz99:
1. component renaming looks good to me

2. This is something I'm still mulling over myself, what the best way to go about this is (including breadcrumbs/button control). One thing to keep in mind is if different people are using the layer we might want to add some flexibility. For example, do we add props to all these components, and set the props in the layout files using the route meta (if people want to make custom layouts, we aren't forcing them to use the route meta this way) or do we get rid of the page meta entirely and only rely on props/store values?. One thought I had was to add props to the components and keep the setBreadcrumbs/setButtonControl utils and then also have a store usePageMetaStore where we would handle all the state utilizing the page meta utils we have. This way we could allow only props or only page meta or only store values to manage the state. Downside is we would have to pass the values in whenever creating a new layout and maybe it's overkill and we should be forcing people to use it in a certain way?. Anyways, I don't see any issue with using the store for strr but we'll want to think about this if pulling it into the layer.

3. 👍

bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-202-kinxn4mf.web.app

bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-202-kinxn4mf.web.app

strr-platform-web/app/interfaces/platform-api.ts
```diff
@@ -4,7 +4,8 @@ export interface ApiAddress {
   addressLineTwo: string
   city: string
   province: string
-  postalCode: string
+  postalCode: string,
+  locationDescription: string
```
deetz99, line 8: Meant to mention this to you. Are we sure we are including the location description? I'm not seeing it in the figma design or the application payload

strr-platform-web/app/middleware/1.redirect.global.ts
deetz99, line 1: Did you try setting a redirect from the nuxt config? Not sure if this works or not.

```
routeRules: {
    '/': { redirect: '/en-CA' }
  },
```

strr-platform-web/app/utils/platform-application.ts
```diff
@@ -5,14 +5,36 @@ export function formatPhoneNumber (phone: ConnectPhone) {
...
 export function formatAddress (add: ConnectAddress): ApiAddress {
   return {
     country: add.country,
     address: add.street,
     addressLineTwo: add.streetAdditional,
     city: add.city,
     province: add.region,
-    postalCode: add.postalCode
+    postalCode: add.postalCode,
+    locationDescription: add.locationDescription
```
deetz99, line 25: We need to confirm locationDescription is being included

strr-platform-web/app/interfaces/platform-api.ts
kialj876: ahh thank you I meant to bring this up to lekshmi. It should be there I think, but I'll double check tomorrow

strr-platform-web/app/middleware/1.redirect.global.ts
kialj876: eh I'm not sure it'll work here unless we add wildcard matching that we'd have to update every time we add a new page

kialj876: I think routeRules config is better for specific redirects (i.e. the page path changed for something and you want peoples bookmarks to still work) vs this is for when people come to our site with any invalid path (which would never happen unless they tried to edit the url manually or typed it out for whatever reason)

deetz99: My thinking was that you were doing the middleware because we don't have a base index page so routeRules would've been fine but if you're using it as a global catch-all then that makes sense the way you did it

strr-platform-web/app/interfaces/platform-api.ts
kialj876: yeah so she'll be adding this in

strr-platform-web/app/middleware/1.redirect.global.ts
kialj876: alright its not really working as expected actually so I'm just going to update the '/' redirect and remove this file

bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-202-kinxn4mf.web.app

bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-202-kinxn4mf.web.app

bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-202-kinxn4mf.web.app

strr-platform-web/app/app.config.ts
deetz99, line 1: Just curious: Isn't merge on by default? Or was this not working properly?

strr-platform-web/app/components/connect/DashboardSection.vue
deetz99, line 1: Nice use of slots, doubt it makes things easier in this context but there is a [USkeleton component](https://ui.nuxt.com/components/skeleton)

strr-platform-web/app/components/connect/DetailsHeader.vue
```diff
@@ -0,0 +1,108 @@
...
+                  @click="btn.action()"
+                />
+              </div>
+            </slot>
+          </div>
+        </div>
+        <dl class="mr-14 space-y-1 text-sm">
+          <template v-for="detail in sideDetails" :key="detail.label">
+            <div>
+              <UTooltip
```
deetz99, line 64: This is a Scott question and I know we aren't adding this now maybe, but how does the user know this field editable? Should we have a little edit icon beside the BN?

strr-platform-web/app/components/connect/typography/H2.vue
deetz99, line 1: Is this correct? 'text-lg' is 18px

strr-platform-web/app/locales/en-CA.ts
deetz99, line 1: Not an issue but I'd like to come up with a naming convention for the translations eventually so everyone is on the same page.

deetz99 (approved):
Just a few minor comments but everything looks great!

strr-platform-web/app/middleware/1.redirect.global.ts
deetz99: I was also thinking it might be good to add a generic 404 page to the layer.

thorwolpert (approved):

strr-platform-web/app/app.config.ts
kialj876: It is, but I like having it set explicitly to remind myself and others that there's the ability to overwrite which may be useful to some projects

strr-platform-web/app/components/connect/DetailsHeader.vue
kialj876: Yeah I was also thinking this needs something like that. Not going to put much more into this though until Scott has a chance to look at it more

strr-platform-web/app/components/connect/typography/H2.vue
kialj876: I think so? We've got H1's as 1.25rem (20px) and H2's as 1.125rem (18px) in many places. The explicit '-32px' / '-24px' were originally from the strr-web project

kialj876: I could be wrong about this for sure though

kialj876: Okay after looking at the other apps I think this is supposed to be 2rem for H1 and 1.5rem for H2, and then we can just override it for other cases where it varies

PR #256, opened 2024-10-31
Platform - UI: Choose account page

deetz99:
*Issue:*
Partial: 
- bcgov/entity#24117

*Description of changes:*
- initial choose account page and middleware

### Todo in future PR:
- add strr product to user account on 'use this account' button click 
- decide if we care about mailing address

Still some todos.

Screenshot 2024-10-31 at 4 52 48 PM
Screenshot 2024-10-31 at 4 53 08 PM
Screenshot 2024-10-31 at 4 53 24 PM
Screenshot 2024-10-31 at 4 54 00 PM

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the strr license (Apache 2.0).

### DIFF
--- a/strr-platform-web/app/middleware/require-premium-account.ts
+++ b/strr-platform-web/app/middleware/require-premium-account.ts
@@ -1,8 +1,9 @@
-export default defineNuxtRouteMiddleware(() => {
+export default defineNuxtRouteMiddleware((to) => {
   const accountStore = useConnectAccountStore()
+  const localePath = useLocalePath()
 
   if (accountStore.currentAccount.accountType !== AccountType.PREMIUM ||
     accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
-    return navigateTo({ path: useLocalePath()('/auth/account/choose-existing') })
+    return navigateTo({ path: localePath('/auth/account/choose-existing'), query: { return: to.fullPath } })
   }
 })
--- a/strr-platform-web/app/middleware/require-premium-account.ts
+++ b/strr-platform-web/app/middleware/require-premium-account.ts
@@ -1,0 +1,8 @@
+export default defineNuxtRouteMiddleware(() => {
+  const accountStore = useConnectAccountStore()
+
+  if (accountStore.currentAccount.accountType !== AccountType.PREMIUM ||
+    accountStore.currentAccount.accountStatus !== AccountStatus.ACTIVE) {
+    return navigateTo({ path: useLocalePath()('/auth/account/choose-existing') })
+  }
+})
--- a/strr-platform-web/app/pages/auth/account/choose-existing.vue
+++ b/strr-platform-web/app/pages/auth/account/choose-existing.vue
@@ -1,0 +1,176 @@
+<script setup lang="ts">
+// import { isoCountriesList } from '~/utils/isoCountriesList'
+const localePath = useLocalePath()
+const { t } = useI18n()
+const isSmallScreen = useMediaQuery('(max-width: 640px)')
+const strrModal = useStrrModals()
+const accountStore = useConnectAccountStore()
+
+useHead({
+  title: t('platform.title.chooseAccount')
+})
+
+definePageMeta({
+  middleware: []
+})
+
+// display premium accounts first
+const sortedAccounts = computed(() => {
+  return accountStore.userAccounts.sort((a, b) => {
+    if (a.accountType === AccountType.PREMIUM && b.accountType !== AccountType.PREMIUM) {
+      return -1
+    } else if (a.accountType !== AccountType.PREMIUM && b.accountType === AccountType.PREMIUM) {
+      return 1
+    }
+    return 0
+  })
+})
+// const sortedAccounts = []
+
+// TODO: how do we want to redirect if user has or does not have any platforms?
+// check if the account has a platform here and either redirect to the dashboard or registration page?
+// or redirect to the dashboard page and if the user doesnt have any platforms they get redirected to the application
+function handleAccountSwitch (id: string) {
+  accountStore.switchCurrentAccount(id)
+  // only navigate if account successfully changed
+  // if (accountStore.currentAccount.id === id) { // TODO: assume account change is always successful ?
+  return navigateTo(localePath('/platform/dashboard'))
+  // }
+}
+
+// TODO: how much do we care about displaying the account mailing address ?
+// if we only display the label, we can use the account list from the connect layer,
+// if we want the mailing address, we'll need to make a separate auth api call
+// maybe this should be added to the layer?
+
+// onMounted(async () => {
+//   const { $authApi } = useNuxtApp()
+
+//   const response = await $authApi('/users/orgs')
+//   console.log(response)
+// })
+</script>
+<template>
+  <div class="space-y-8 py-8 sm:py-10">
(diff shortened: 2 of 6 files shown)


bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-256-0uqs5r0b.web.app

deetz99:
I think process.env.NUXT_BASE_URL may not be set correctly in in the gcbrun

strr-platform-web/app/pages/auth/account/choose-existing.vue
```diff
@@ -0,0 +1,182 @@
...
+  // if (accountStore.currentAccount.id === id) { // TODO: assume account change is always successful ?
+
+  if (route.query.return) {
+    return navigateTo(route.query.return as string)
+  } else {
+    return navigateTo(localePath('/platform/dashboard'))
+  }
+}
+
+// TODO: how much do we care about displaying the account mailing address ?
```
kialj876, line 47: I don't think we need to display this, but we probably want to display the branch info? And I think the branch info should also be in the header dropdown, so worth adding to the layer

kialj876:
> I think process.env.NUXT_BASE_URL may not be set correctly in in the gcbrun

Hmm yeah that is a tricky one to fix because the url changes each time. I'll have a look and see if we have access to the pr url before the env is set

strr-platform-web/app/pages/auth/account/choose-existing.vue
```diff
@@ -0,0 +1,182 @@
...
+      return -1
+    } else if (a.accountType !== AccountType.PREMIUM && b.accountType === AccountType.PREMIUM) {
+      return 1
+    }
+    return 0
+  })
+})
+// const sortedAccounts = []
+
+// TODO: how do we want to redirect if the user has/does not have any platforms?
```
kialj876, line 30: always to the dashboard. If they don't have any platforms then they'll have the 'begin application' todo in there

strr-platform-web/app/pages/auth/account/choose-existing.vue
```diff
@@ -0,0 +1,182 @@
...
+// check if the account has a platform here and either redirect to the dashboard or registration page?
+// or redirect to the dashboard page and if the user doesnt have any platforms the dashboard can handle the redirect ?
+function handleAccountSwitch (id: string) {
+  const route = useRoute()
+  accountStore.switchCurrentAccount(id)
+  // TODO: add Strr product to org product list
+  // only navigate if account successfully changed
+  // if (accountStore.currentAccount.id === id) { // TODO: assume account change is always successful ?
```
kialj876, line 38: I think we can assume it is successful as it isn't dependent on a network call or anything like that. Sounds like we don't need to add the product for now (it works for them without the product)

strr-platform-web/app/middleware/require-premium-account.ts
kialj876, line 1: I am thinking we probably also want to check that they actually selected an account explicitly. Right now we just default to the first one in their list, but if they have multiple we talked about forcing them to choose which one to use so that they don't accidentally file things under the wrong account.

To do this I was thinking we could add a flag in session storage or cookie like 'userSelectedAccount' or something like that. I don't think we can save it as a store value (unless piniaPersistedstate works)  because we don't want them to have to reselect when they navigate to other nuxt apps within bcregistries

lmk what you think -- we can add this in later

kialj876 (approved):
Nice work 👍

strr-platform-web/app/middleware/require-premium-account.ts
deetz99: This should still fail if there's no account selected and redirect to the choose account page. Do we want to redirect them to this page if they have more than 1 premium account? Once people are comfortable with the system this might be kind of annoying instead of just going to the dashboard and picking whatever account they want to use.

strr-platform-web/app/pages/auth/account/choose-existing.vue
forgeuxGH5: No address - branch is required

strr-platform-web/app/pages/auth/account/choose-existing.vue
forgeuxGH5: short-term (for now) we can redirect to the dashboard - longer term we want to direct as in the figjam flow. Zero platforms would launch the registration application. 1 platform would go to the platform dashboard. Future looking, we will need to handle multiple "objects" when we get to Strata Hotels, so 2 or more "objects" (e.g., strata hotels) would redirect to the "My STRR Dashboard (BRD) (the Host team is working on this too).

strr-platform-web/app/middleware/require-premium-account.ts
forgeuxGH5: For now, this page should always be displayed after they enter their login credentials / are successfully logged in. They have to select an account or create a new account to proceed. Long term, we want to direct them based on logic if they have 0,1, or 2 or more accounts.

forgeuxGH5:
What's an inactive account? How is this status determined?

deetz99:
@forgeuxGH5 

I have it so the badge will show up if the account is anything other than 'ACTIVE', this is the account object I have for now.

```
    {
        "accountStatus": "ACTIVE",
        "accountType": "BASIC",
        "id": 3270,
        "label": "Test AR Account 2",
        "productSettings": "/account/3270/restricted-product",
        "type": "ACCOUNT",
        "urlorigin": "https://dev.bcregistry.ca/business/auth",
        "urlpath": "/account/3270/settings"
    },
```

The account can have any status below:

```
export enum AccountStatus {
  ACTIVE = 'ACTIVE',
  INACTIVE = 'INACTIVE',
  NSF_SUSPENDED = 'NSF_SUSPENDED',
  PENDING_STAFF_REVIEW = 'PENDING_STAFF_REVIEW',
  PENDING_ACTIVATION = 'PENDING_ACTIVATION',
  REJECTED = 'REJECTED',
  SUSPENDED = 'SUSPENDED'
}
```

deetz99:
@forgeuxGH5 

Is branch name required for MVP? We'll need to make a change to the core layer if so. I will merge this PR just to keep things moving but let me know if there's any changes you want.

forgeuxGH5:
for the badge let's use the following Registries style colours to be WCAG AA conforming:
Text: #495057 (bold, all caps would be nice)
Background fill: FFF7E3

forgeuxGH5:
Can we determine the payment type of the Premium account? My understanding is they need a Premium account with PAD - right now existing account could have a Premium account with either BCOL payment or PAD, and platforms can't use BCOL payments, so we want to prevent them selecting Premium Accounts with a BCOL payment type. Or am I wrong about the BCOL payments?

thorwolpert:
> @forgeuxGH5
> Can we determine the payment type of the Premium account? My understanding is they need a Premium account with PAD - right now existing accounts could have a Premium account with either BCOL payment or PAD, and platforms can't use BCOL payments, so we want to prevent them selecting Premium Accounts with a BCOL payment type. Or am I wrong about the BCOL payments?
It'll work if they have a BCOL account, it's unlikely they'll get one approved so I think it makes that a non-issue

strr-platform-web/app/middleware/require-premium-account.ts
forgeuxGH5: Yeah - we always want to go through the choose account page after login (for now)

PR #259, opened 2024-11-01
UI Base web layer

kialj876:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*


kialj876:
broke something in the platform ui, will fix on Monday

strr-base-web/app/pages/auth/login.vue
```diff
@@ -1,14 +1,13 @@
 <script setup lang="ts">
 const { t, locale } = useI18n()
 const keycloak = useKeycloak()
-const redirectUrl = useRuntimeConfig().public.baseUrl + locale.value + '/platform/dashboard'
-const logoutUrl = useRuntimeConfig().public.baseUrl + locale.value + '/auth/login'
+const redirectUrl = useRuntimeConfig().public.baseUrl + locale.value + '/auth/login'
```
deetz99, line 4: We can remove explicitly setting the redirect url here. Keycloak will redirect to the same url it was triggered from if there's no redirect url set.

deetz99 (approved):
Looks good! 👍

PR #263, opened 2024-11-04
Strr base layer updates

kialj876:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*


strr-base-web/package.json
kialj876, line 1: NOTE: removed deps are coming from the @daxiom/nuxt-core-layer-test already, moved 'nuxt' from devDependency to dependency (it is a devDependency in @daxiom/nuxt-core-layer-test)

bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-263-qokowyf5.web.app

strr-platform-web/app/components/connect/form/phoneNumber/CountryCode.vue
```diff
@@ -0,0 +1,121 @@
+<script setup lang="ts">
+// NOTE: country-codes-list fails if not overridden (even though it is in the repo base layer)
```
kialj876, line 2: FYI - this was weird and I think it's due to a combination of the moduel 'country-codes-list' and the way Nuxt imports from github repo extensions (it is different than when it extends from npm packages). Will experiment more with this later, but for now we need a copy of ConnectFormPhoneNumberCountryCode in each layer. I think this will go away once its coming from an npm package vs git repo

kialj876:
> Temporary Url for review: https://strr-platform-dev--pr-263-qokowyf5.web.app

NOTE: this has the ldarkly redirect issue because it is pulling the dependency from the checked in repo code. I have verified that the new code inside this PR fixes that issue. Still not sure why the PR build is not picking up the env vars for ldarkly and address complete

bcregistry-sre:
Temporary Url for review: https://strr-platform-dev--pr-263-qokowyf5.web.app

strr-base-web/package.json
deetz99: I don't think the devDependencies get carried forward from the @daxiom layer.

In this case its just vue and vue-router but I think nuxt should handle this already so it should be a non-issue hopefully

strr-platform-web/nuxt.config.ts
```diff
@@ -8,7 +8,8 @@ export default defineNuxtConfig({
   },
 
   routeRules: {
-    '/': { redirect: '/en-CA/platform/dashboard' }
+    '/': { redirect: '/en-CA/platform/dashboard' },
+    '/en-CA': { redirect: '/en-CA/platform/dashboard' }
```
deetz99, line 12: might have a conflict here, I also added `'/fr-CA': { redirect: '/fr-CA/platform/dashboard' }`

deetz99 (approved):
Nice work 👍

strr-base-web/package.json
kialj876: yeah they do get carried from nuxt and it handles the versioning etc. so I think its actually better to not include it explicitly

PR #291, opened 2024-11-09
Platform/Strata - UI: update breadcrumbs

deetz99:
Partial: bcGov/entity#24327

Using a store we can optionally add the account id and it will stay reactive when the user switches accounts, let me know your thoughts. This would get moved into the core layer if we're happy with it. Will need to update all other pages if this is the route we want to go.

Actually, maybe we keep the breadcrumbs in the route meta and we still add the appendAccountId property. Then in the breadcrumb component we can just add the account id using computed the way I did it in the store. This would allow us to still use the setBreadcrumbs auto imported util.

This is the old breadcrumb computed:
```
const breadcrumbs = computed<BreadcrumbLink[]>(() => {
  const route = useRoute()

  if (route.meta.breadcrumbs) {
    return route.meta.breadcrumbs
  } else {
    return [{ label: t('ConnectBreadcrumb.default') }]
  }
})
```

We could add this to it and keep using the route meta instead of a store:
```
if (bc.appendAccountId && accountStore.currentAccount.id) {
        return {
          ...bc,
          to: appendUrlParam(bc.to as string, 'accountid', accountStore.currentAccount.id)
        }
      }
```


kialj876 (commented):
Yeah I actually do like what you've done by making it a store here more than the 'meta way' because we were  sort of using it as a pseudo store combined with the util fn.

I also like what you mentioned for updating the computed if we don't go with this

bcregistry-sre:
Temporary Url for review: https://strr-strata-hotel-dev--pr-291-6ym6w9k6.web.app

strr-strata-web/app/pages/comingSoon.vue
deetz99, line 1: moved to strr-base

kialj876 (approved):
Nice work, looks good to me 👍

PR #293, opened 2024-11-11
UI - Rental Unit Address Breakout

shaangill025:
*Issue:*

- https://github.com/bcgov/entity/issues/24125

*Description of changes:*


strr-web/components/bcros/form-section/property/Address.vue
```diff
@@ -22,59 +22,118 @@
...
+          class="d:pr-[16px] flex-grow"
+          :error="streetNameError"
+        >
+          <UInput
+            :id="streetNameId"
+            v-model="streetName"
+            :aria-label="t('createAccount.contactForm.streetName')"
+            :placeholder="t('createAccount.contactForm.streetName')"
+            @input="
+              onAddressInput(true); emit('resetFieldError', [
```
dimak1, line 56: Do we need to call `onAddressInput()` ? It might be tricky with my validation refactoring.

strr-web/components/bcros/form-section/property/Address.vue
```diff
@@ -22,59 +22,118 @@
...
-            :id="id"
-            v-model="address"
-            :placeholder="t('createAccount.contactForm.address')"
-            aria-label="address"
-            @input="onAddressInput"
-            @click="addressComplete()"
+            v-model="unitNumber"
+            :aria-label="t('createAccount.contactForm.unitNumberRequired')"
+            :placeholder="unitNumberPlaceholder"
+            @input="emit('resetFieldError', ['unitNumber'])"
```
dimak1, line 74: I noticed that this is different from earlier implementation, where `resetFieldError` was emitting a string (and not array). Something changed, you want to reset multiple fields at once?

strr-web/components/bcros/form-section/property/Address.vue
```diff
@@ -84,57 +143,107 @@
...
 }
 
 const {
-  id,
+  streetNumberId,
+  streetNameId,
   defaultCountryIso2,
   enableAddressComplete,
-  addressNotInBC
+  addressNotInBC,
```
dimak1, line 188: Why not pass a single error object like before, instead of one by one... If you feel like doing small updated, you can refactor, otherwise leave to me for my validation updates.

strr-web/components/bcros/form-section/property/Form.vue
```diff
@@ -69,20 +88,31 @@ const { isComplete } = defineProps<{
...
+  if (activeAddressField.value === 'propertyAddressStreetNumber' ||
+      activeAddressField.value === 'propertyAddressStreetName') {
+    return formState.propertyDetails
+  }
+  return null
+}
 
 watch(canadaPostAddress, (newAddress) => {
-  if (newAddress) {
+  const activeAddressState = getActiveAddressState()
```
dimak1, line 105: Add a type for safety, either to the function or here.

strr-web/components/bcros/form-section/property/Form.vue
```diff
@@ -273,6 +303,80 @@ const validateNumberOfRoomsForRent = () => {
...
+  provinceError.value = error ? error.message : ''
+}
+
+const validatePostalCode = () => {
+  const parsed = propertyDetailsSchema.safeParse(formState.propertyDetails).error?.errors
+  const error = parsed?.find(error => error.path.includes('postalCode'))
+  postalCodeError.value = error ? error.message : ''
+}
+
+const resetFieldError = (errorTypes: string[]) => {
```
dimak1, line 356: Ok I see it now... can reset multiple fields...

shaangill025: Yes, with `streetNumber` and `streetName` [which triggers address complete], I want to reset errors for `city`, `postal code`, etc.

strr-web/components/bcros/form-section/property/Address.vue
shaangill025: When a new address is being inputted and autocomplete triggers then previously set fields like `city`, `province` and other need to be reset. This is essentially what's happening with `onAddressInput`.

shaangill025:
Comments to address from #292 
 > It is very similar to `CanadaPostAddressWithStreetAttributesI` interface. Why have two? Could we just extend the existing interface to include additional fields as optional?
> Having multiple address interfaces is confusing and adds a tech debt, because it is hard to know which one to use.


> It looks like the values belong to a content en.json and don't look like a value map to me. Enums usually self-mapping. Here is a sample:
> 
> ```
> export enum PropertyTypeE {
> SECONDARY_SUITE = 'SECONDARY_SUITE'
> }
> 
> const PropertyTypeLabels = {
>   [PropertyTypeE.SECONDARY_SUITE]: 'Secondary Suite'
> }
> ```
> 
> I see that further in the code, the values themselves map to content strings (meaning they don't have to be fully spelled out here).


> Probably should be `fetchedRegistration?.unitAddress` for checking if actually unit address exists and not only registration


> Thanks for refactoring the Property Details validation. Could you please add some comments to explain what is going on here with chaining transform/pipe? Will be good for future readability and knowledge sharing. Thanks

strr-web/components/bcros/form-section/property/Address.vue
dimak1: Hmm, shouldn't it be cleared only after I select an address from auto-complete? What if I change my mind and don't select anything, then those fields are gone?

strr-web/components/bcros/form-section/property/Address.vue
dimak1: If this is what the requirement is, I'm good with it, thanks for clrifying.

strr-web/components/bcros/form-section/property/Form.vue
shaangill025: Fixed

strr-web/components/bcros/form-section/property/Address.vue
shaangill025: Fields are not being cleared now.

dimak1:
I started to review the PR but please run it thru UXA as well, thanks @shaangill025

dimak1 (approved):
LG, good job!

PR #346, opened 2024-12-02
Update Auto-Approval page for Examiner

dimak1:
*Issue:*

Going to the Auto-Approval Details page was failing because the application number was not available to fetch the details.

*Description of changes:*
- Update Auto-Approval page to fetch details based on `applicationNumber`
- Update AutoApprovalDataI interface as per latest API updates


strr-web/pages/application-details/[id]/auto-approval/index.vue
```diff
@@ -52,16 +52,22 @@ const tAutoApproval = (translationKey: string) => t(`autoApproval.${translationK
...
 const { setupBreadcrumbData } = useBreadcrumb()
 
-const applicationNumber = route.params.id.toString()
+const registrationId = parseInt(route.params.id.toString())
 
-const application = await getApplication(applicationNumber)
-const data: AutoApprovalDataI[] = await getAutoApproval(applicationNumber) || {} as AutoApprovalDataI[]
-const applicationDetails = application.registration as HostApplicationDetailsI
+// Fetch applications and find application number based on registration id from params
+const { applications } = await useApplications().getApplications()
```
Comment (dimak1) [inline strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: @kris-daxiom there will be an extra API call to fetch applications, because on this page the `applicationNumber` is not available otherwise to get the auto-approval details. So instead of making changes to API, I went ahead with this change (for now, until we reworks this page later on).

Comment (kialj876) [reply strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: You can avoid this if you open this page via the application number. Is there a reason you're using the registration number instead of the application number to open this page?

kialj876 (approved):
I approved so that you're unblocked for your demo todat, but see my comment below (maybe you can update it later)

strr-web/pages/application-details/[id]/auto-approval/index.vue
Comment (kialj876) [reply strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: Also I think if there are > 50 results the application you're looking for may not be in the response?

Comment (dimak1) [reply strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: @kialj876 This is a Registration Details page, so we need to get Registration info by fetching using the `reg id`. API needs `app number` to get the Auto-Approval info. We wanted to consolidate Application and Registration Details into one page, but didn't have enough time. I hope to cleanup and do some API updates once we rebuild the page.

Comment (dimak1) [reply strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: @kialj876 The one application we want _should be_ in the response because before any Registration exists, there should be an Application. So this page is technically a step 2 and application is step 1, so should exist.

Comment (kialj876) [reply strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: yeah but the api has paging so you'll only get a max of 50 results. So if there's 500 applications and the one you're looking for isn't in the first 50 then you won't get it

Comment (kialj876) [reply strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: The way the api expects you to do this is by 1st getting the application, then 2nd get the registration for that application (if it exists then application will have the registration number for you)

Comment (dimak1) [reply strr-web/pages/application-details/[id]/auto-approval/index.vue:61]: Yea i see what you mean, there are some limitations, we'll chat offline.

PR #352, opened 2024-12-04
fixes

kris-daxiom:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*

- Changes to add business as a co-host
- Added new value to the ownership type enum
- Added new document types
- Fixed the dob format issue in the registration response
- Added phone country code for hosts
- Updated the payment service code to override quantity only if the number of rooms shared is greater than zero
- Added tests and updated a few existing tests


kialj876 (approved):
Nice work, thanks for the quick update!

thorwolpert (approved):

PR #354, opened 2024-12-05
Hosts - Unit address input autocomplete or manual entry

deetz99:
*Issue:*

- bcgov/entity/issues/24738

*Description of changes:*
- separate input to use either autocomplete OR manual input
- custom residential address component locked to BC/Canada
- update residential address schema to match autocomplete/manual entry option and make street number/name optional if line 2 has a value


kialj876 (approved):
Nice work 👍

PR #355, opened 2024-12-05
24702 hosts review submit

kialj876:
*Issue:*

- bcgov/entity/issues/24702

*Description of changes:*


strr-host-pm-web/app/stores/document.ts
```diff
@@ -228,7 +232,8 @@ export const useDocumentStore = defineStore('host/document', () => {
         method: 'POST',
         body: formData
       })
-
+      // api doesn't give documentType back
+      res.documentType = uiDoc.type
```
deetz99, line 236: Is this necessary if the apiDocuments computed above adds the documentType already? or maybe the computed could be simplified to just 

```
const apiDocuments = computed(() => storedDocuments.value.map(item => item.apiDoc))
```

Also my bad with the documentType property name

strr-host-pm-web/app/stores/hostApplication.ts
```diff
@@ -1,55 +1,96 @@
...
+        unitDetails: formatHostUnitDetailsAPI(propertyStore.unitDetails, propertyStore.blInfo),
+        documents: documentStore.apiDocuments,
+        principalResidence: {
+          // TODO: confirm mapping - assuming this is if PR is required or not?
+          isPrincipalResidence: !reqStore.prRequirements.isPropertyPrExempt,
+          nonPrincipalOption: reqStore.prRequirements.prExemptionReason || '',
+          agreedToRentalAct: userConfirmation.value.agreedToRentalAct,
+          agreedToSubmit: userConfirmation.value.agreedToSubmit,
+          // TODO: No idea what this is supposed to be?
+          specifiedServiceProvider: ''
```
deetz99, line 60: This was the 'other service provider' option from before.

If you selected 'other' for the reason youre pr exempt, you had to give the other reason. We don't have other as an exemption value or the `specifiedServiceProvider` in the new design.

strr-host-pm-web/app/stores/hostApplication.ts
```diff
@@ -1,55 +1,96 @@
...
+        ...(propertyManger
+          ? { propertyManager: formatOwnerPropertyManagerAPI(propertyManger) }
+          : {}
+        ),
+        unitAddress: formatHostUnitAddressApi(propertyStore.unitAddress.address),
+        unitDetails: formatHostUnitDetailsAPI(propertyStore.unitDetails, propertyStore.blInfo),
+        documents: documentStore.apiDocuments,
+        principalResidence: {
+          // TODO: confirm mapping - assuming this is if PR is required or not?
+          isPrincipalResidence: !reqStore.prRequirements.isPropertyPrExempt,
```
deetz99, line 55: Would need to confirm but I think this is the rental unit setup type. rentalUnitSetupType === RentalUnitSetupType.WHOLE_PRINCIPAL_RESIDENCE

I'm just not sure if it needs to be the whole unit or if a room in the pr is okay too.

strr-host-pm-web/app/stores/hostOwner.ts
```diff
@@ -89,11 +89,27 @@ export const useHostOwnerStore = defineStore('host/owner', () => {
...
+
+    if (returnBool) {
+      return success
+    }
+    return [{
+      formId: 'host-owners',
+      success,
+      errors: [
+        ...(!hasCompParty.value
+          ? [{ message: 'Missing completing party', code: 'custom', path: ['owners.completingParty'] } as z.ZodIssue]
```
deetz99, line 106: Do we need i18n here?

deetz99 (approved):
Few comments but nothing major, looks great 👍

strr-host-pm-web/nuxt.config.ts
```diff
@@ -61,7 +61,8 @@ export default defineNuxtConfig({
       housingLearnMoreUrl: process.env.NUXT_HOUSING_LEARN_MORE_URL,
       housingAllRulesUrl: process.env.NUXT_HOUSING_ALL_RULES_URL,
       housingRequiredDocsUrl: process.env.NUXT_HOUSING_REQUIRED_DOCS_URL,
-      housingProofOfPrUrl: process.env.NUXT_HOUSING_PROOF_OF_PR_URL
+      housingProofOfPrUrl: process.env.NUXT_HOUSING_PROOF_OF_PR_URL,
+      hostTacUrl: process.env.NUXT_HOST_TAC_URL
```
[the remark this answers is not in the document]
deetz99: Actually the terms and conditions link is different it seems per app so nevermind

strr-host-pm-web/app/stores/hostApplication.ts
kialj876: I'll confirm this with lekshmi today

strr-host-pm-web/app/stores/document.ts
kialj876: Nice catch, I will update the computed since the api validation requires 'documentType' so it should be with the 'apiDoc' always (the api response when we get the application will include 'documentType' in the apiDoc so this will keep it the same)

strr-host-pm-web/app/stores/hostOwner.ts
kialj876: I don't think its necessary because they aren't shown in the UI (would only ever see them in the logs)

strr-host-pm-web/app/stores/hostApplication.ts
kialj876: haven't been able to get ahold of her so I'm going to merge and then will update as needed

PR #357, opened 2024-12-06
Initial setup for new Examiners app

dimak1:
*Issue:*

- bcgov/entity/issues/24717

*Description of changes:*
- Extend base layer and setup new Examiners app
- Extend from Base, Host, Platform, Strata to leverage existing interfaces (might need to re-think this later)
- Auth works using only IDIR
- Fetch existing Application and Registrations

Screenshot 2024-12-06 at 08 07 24


avni-work:
hey dima, do you need a ci-cd for this? if yes, when it is ready can you please throw me a ticket?

dimak1:
@avni-work yes, please use #24717, I've added a note about it.

strr-examiner-web/nuxt.config.ts
```diff
@@ -0,0 +1,90 @@
...
+    ],
+    strategy: 'prefix',
+    lazy: true,
+    langDir: 'locales',
+    defaultLocale: 'en-CA',
+    detectBrowserLanguage: false,
+    vueI18n: './i18n.config.ts'
+  },
+
+  extends: [
```
deetz99, line 41: Keep in mind extends will bring over everything (pages/components/utils/composables/etc), so all the pages that are in the strata/platform/host app will be available in the examiner view this way. 

You can ignore certain paths from being prerendered but in this case we should only need to extend the base layer. Mostly everything thats reusable has been added to the base layer. If there are specific things you'd like from the other apps we can look into moving those to the base layer.

PR #364, opened 2024-12-09
Hosts dashboard/form fix

deetz99:
*Issue:*

- bcgov/entity/issues/24809
- bcgov/entity/issues/24808

*Description of changes:*

### DIFF
--- a/strr-host-pm-web/app/stores/propertyRequirements.ts
+++ b/strr-host-pm-web/app/stores/propertyRequirements.ts
@@ -78,7 +78,7 @@ export const usePropertyReqStore = defineStore('property/requirements', () => {
       propertyReqs.value = res
 
       // TODO: confirm these are the only scenarios we want to open the form without further user interaction
-      if (!res.isStrProhibited || !!res.isStraaExempt) {
+      if (!res.isStrProhibited && !res.isStraaExempt) {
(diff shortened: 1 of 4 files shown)


strr-host-pm-web/app/stores/propertyRequirements.ts
```diff
@@ -78,7 +78,7 @@ export const usePropertyReqStore = defineStore('property/requirements', () => {
       propertyReqs.value = res
 
       // TODO: confirm these are the only scenarios we want to open the form without further user interaction
-      if (!res.isStrProhibited || !!res.isStraaExempt) {
+      if (!res.isStrProhibited && !res.isStraaExempt) {
```
kialj876, line 81: My understanding of these rules is a little hazy, can you confirm that this works as expected with the 'Add supporting information' tab and 'review / confirm' tab? Both of those use showUnitDetailsForm when deciding whether or not to show 'no documents required' vs 'incomplete'

kialj876 (approved):
Looks good, one verify question

strr-host-pm-web/app/stores/propertyRequirements.ts
deetz99: Sure thing

PR #436, opened 2025-01-06
Fix Examiner Dash and Details page

dimak1:
*Issue:*

- bcgov/entity/issues/25145

*Description of changes:*
- Fix Examiner Dash and Details page

### DIFF
--- a/strr-web/pages/application-details/[id]/index.vue
+++ b/strr-web/pages/application-details/[id]/index.vue
@@ -187,34 +187,40 @@
         </div>
 
         <!-- Principal Residence -->
-        <div class="mt-10" data-test-id="principal-residence">
+        <div
+          v-if="applicationDetails.strRequirements"
+          class="mt-10"
+          data-test-id="principal-residence"
+        >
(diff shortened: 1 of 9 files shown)


strr-web/pages/application-details/[id]/index.vue
Comment (deetz99) [inline strr-web/pages/application-details/[id]/index.vue:1]: Missing some i18n on the titles. Also what happens if there are no strRequirements? 

If the user enters an address that cant be found in the data portal, we still let them submit. In this case the strRequirements gets submitted as an empty object but the application could still be valid.

Comment (dimak1) [reply strr-web/pages/application-details/[id]/index.vue:1]: I figured that this app is going away and translations not needed (but happy to add the titles to i18n file).

For the strRequirements, I have a check on line 191. If they exists, I basically display them as is (Fiona was approved), Should've included this in the description.

Screenshot 2025-01-06 at 10 31 06

Comment (deetz99) [reply strr-web/pages/application-details/[id]/index.vue:1]: Works for me

deetz99 (approved):

PR #444, opened 2025-01-09
Add Platform and Strata Details pages

dimak1:
*Issue:*

- bcgov/entity/issues/24933

*Description of changes:*
- Add Host, Platform, Strata Details pages via their own components
- Add interfaces so we don't have to import via layers
- Update header, content fix

Screenshot 2025-01-09 at 08 44 48

Screenshot 2025-01-09 at 08 27 20

Screenshot 2025-01-09 at 08 49 09


strr-examiner-web/app/components/HostDetailsView.vue
kialj876, line 1: What do you think about 1 component or a template component for DetailsView to reduce the duplication? Looks like these DetailsView comps are mostly the same. And i18n for all the labels etc.?

deetz99:
Just wondering if it would be better to move all the interfaces and enums to the base layer, this way it's easier to update whenever there is a change.

strr-examiner-web/app/components/HostDetailsView.vue
dimak1: I was thinking about the same, having reusable component. It's just we don't have any designs yet (for platform/strata), so I have no idea how details pages will look. So i created separate placeholder components for now.

dimak1:
> Just wondering if it would be better to move all the interfaces and enums to the base layer, this way it's easier to update whenever there is a change.

@deetz99 yea that would be great. This was a initial stab at the interfaces, to see how much we actually need, hence kept all the names the same for easier refactoring.

deetz99:
Is this the only design available right now? https://www.figma.com/design/gzey7iitwuZ39IYMtNX3tg/Platform-Dashboard?node-id=1518-500&t=8fiuJlZxzcLbWPi2-1

strr-examiner-web/app/components/HostDetailsView.vue
kialj876: gotcha, sounds good 👍

kialj876 (approved):

dimak1:
@deetz99 yes this is only design (Hosts) available now.

deetz99 (approved):

PR #465, opened 2025-01-16
API - update draft application date every time it is updated

kialj876:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*
- suggested change for drafts

### DIFF
--- a/strr-api/src/strr_api/resources/application.py
+++ b/strr-api/src/strr_api/resources/application.py
@@ -37,6 +37,7 @@ STRR Application Resource.
 """
 
 import logging
+from datetime import datetime, timezone
 from http import HTTPStatus
 from io import BytesIO
 from typing import Optional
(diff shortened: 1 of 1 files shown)


strr-api/src/strr_api/resources/application.py
```diff
@@ -125,6 +126,8 @@ def create_application(application_number: Optional[str] = None):
                     message=ErrorMessage.APPLICATION_NOT_MODIFIABLE.value,
                     http_status=HTTPStatus.BAD_REQUEST,
                 )
+            # update application date
+            application.application_date = datetime.now(timezone.utc)
```
kialj876, line 130: alternatively we could have this be inside `if not is_draft:` so that it only updates when the application is completed.

You may also prefer this is moved to the application service, but I thought this might be simpler / its easy to know it only affects drafts

PR #486, opened 2025-01-22
Strata Details page - Part 2/2

dimak1:
*Issue:*

- bcgov/entity/issues/25191

*Description of changes:*
- Add Expansion panels for Strata details

Screenshot 2025-01-21 at 16 00 12


deetz99 (approved):
Looks good to me 👍  

The only comment would be that because the ExpansionTemplate component is handling the close fn, you don't need the emits in the strata expansions.

Nice work!

dimak1:
@deetz99 good catch, thanks, I've updated the composable

deetz99:
> @deetz99 good catch, thanks, I've updated the composable

Looks good 👍

PR #493, opened 2025-01-24
UI: Examiner table filters

deetz99:
*Issue:*

- bcgov/entity/issues/25369

*Description of changes:*
- add ui table filter items
- didn't fully bother with i18n and a11y stuff as these will probably change a bit
- USelectMenu ui prop not working correctly so had to override with :deep selector
- currently only only application number, reg type and status filters work
- currently cannot have more than 1 reg type or status active 
- added the search endpoint to the application number filter input, this will only work if the reg type filter is empty (it will also match on other fields than the application number)
- sort also not working yet

Screenshot 2025-01-24 at 11 36 39 AM


strr-examiner-web/nuxt.config.ts
deetz99, line 1: my hmr was pretty broken, adding this seemed to help

strr-examiner-web/app/stores/examiner.ts
```diff
@@ -46,11 +60,32 @@ export const useExaminerStore = defineStore('strr/examiner-store', () => {
...
     approveApplication,
     rejectApplication,
     getNextApplication,
     getDocument,
-    openDocInNewTab
+    openDocInNewTab,
+    resetFilters
   }
-})
+}, { persist: true }) // will persist data in session storage
```
dimak1, line 91: 👍🏻

dimak1 (commented):
Looks great!

dimak1 (approved):

kialj876 (approved):
Great work!

PR #575, opened 2025-02-27
add uat to clouddeploy and cd files

avni-work:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*

### DIFF
--- a/queue_services/strr-email/devops/gcp/clouddeploy.yaml
+++ b/queue_services/strr-email/devops/gcp/clouddeploy.yaml
@@ -31,6 +31,19 @@ serialPipeline:
       container-name: "strr-email-test"
       cloudsql-instances: "bcrbk9-test:northamerica-northeast1:strr-db-test"
       service-account: "sa-api@bcrbk9-test.iam.gserviceaccount.com"
+ - targetId: bcrbk9-test
+   profiles: [uat]
+   strategy:
+    standard:
+      verify: false
+   deployParameters:
+   - values:
+      deploy-env: "uat"
+      deploy-project-id: "bcrbk9-test"
+      service-name: "strr-email-uat"
+      container-name: "strr-email-uat"
+      cloudsql-instances: "bcrbk9-test:northamerica-northeast1:strr-db-test"
+      service-account: "sa-api@bcrbk9-test.iam.gserviceaccount.com"
  - targetId: bcrbk9-sandbox
    profiles: [sandbox]
    strategy:
--- a/queue_services/strr-pay/devops/gcp/clouddeploy.yaml
+++ b/queue_services/strr-pay/devops/gcp/clouddeploy.yaml
@@ -31,6 +31,19 @@ serialPipeline:
       container-name: "strr-pay-test"
       cloudsql-instances: "bcrbk9-test:northamerica-northeast1:strr-db-test"
       service-account: "sa-api@bcrbk9-test.iam.gserviceaccount.com"
+ - targetId: bcrbk9-test
+   profiles: [uat]
+   strategy:
+    standard:
+      verify: false
+   deployParameters:
+   - values:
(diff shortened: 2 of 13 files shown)


queue_services/strr-pay/devops/gcp/clouddeploy.yaml
```diff
@@ -31,6 +31,19 @@ serialPipeline:
...
       cloudsql-instances: "bcrbk9-test:northamerica-northeast1:strr-db-test"
       service-account: "sa-api@bcrbk9-test.iam.gserviceaccount.com"
+ - targetId: bcrbk9-test
+   profiles: [uat]
+   strategy:
+    standard:
+      verify: false
+   deployParameters:
+   - values:
+      deploy-env: "uat"
```
bolyachevets, line 41: deploy-env value is used here: https://github.com/bcgov/bcregistry-sre/blob/76332c3c697589f65fa74febe85aae49d9cec1d5/.github/actions/backend-deploy/files/k8s/service.template.yaml#L46 The value of DEPLOYMENT_ENV is used for POD_NAMESPACE variable, it looks like only 'migration' value is used: https://github.com/bcgov/STRR/blob/1f4e45e44f3b5671f5f645cac3cbffd51efe5ffa/strr-api/src/strr_api/__init__.py#L68. Maybe need to double check that  this is intended and how environment specific configs are initialized

queue_services/strr-email/devops/gcp/clouddeploy.yaml
```diff
@@ -31,6 +31,19 @@ serialPipeline:
...
+   strategy:
+    standard:
+      verify: false
+   deployParameters:
+   - values:
+      deploy-env: "uat"
+      deploy-project-id: "bcrbk9-test"
+      service-name: "strr-email-uat"
+      container-name: "strr-email-uat"
+      cloudsql-instances: "bcrbk9-test:northamerica-northeast1:strr-db-test"
```
bolyachevets, line 45: are you going to connect to the same db? (maybe same instance is ok, but would we need a different db in that instance? - checked currently only strr-test db lives in bcrbk9-test:northamerica-northeast1:strr-db-test)

avni-work: yes same db

strr-api/devops/gcp/clouddeploy.yaml
```diff
@@ -45,6 +45,19 @@ serialPipeline:
...
+   strategy:
+    standard:
+      verify: false
+   deployParameters:
+   - values:
+      deploy-env: "uat"
+      deploy-project-id: "bcrbk9-test"
+      service-name: "strr-api-uat"
+      container-name: "strr-api-uat"
+      cloudsql-instances: "bcrbk9-test:northamerica-northeast1:strr-db-test"
```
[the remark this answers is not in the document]
avni-work: yes same db, but different code will be deployed.
would there be a difference in uat api code vs test api code?

queue_services/strr-pay/devops/gcp/clouddeploy.yaml
avni-work: I think your concern is valid!

we could do the following in the __init__.py. would you think that would mitigate this?

if (app.config.get("POD_NAMESPACE", "production") == "migration" or 
    app.config.get("DEPLOYMENT_ENV", "production") == "migration"):
    logger.info("Running in migration mode")
    Migrate(app, db)

queue_services/strr-email/devops/gcp/clouddeploy.yaml
pwei1018: may keep using the value 'test' of deploy-env without change the code.

deploy-env: "test"

PR #600, opened 2025-03-11
API - Remove REJECT examiner action - FULL_REVIEW application status

shaangill025:
*Issue:*

- https://github.com/bcgov/entity/issues/26241

*Description of changes:*

### DIFF
--- a/strr-api/pyproject.toml
+++ b/strr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "strr-api"
-version = "0.0.51"
+version = "0.0.52"
 description = ""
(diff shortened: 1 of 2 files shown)


dimak1 (approved):

dimak1:
@shaangill025 should the `useExaminerRoute.ts` needs to be updated to remove REJECT action as well?

shaangill025:
> @shaangill025 should the `useExaminerRoute.ts` needs to be updated to remove REJECT action as well?

Let's keep it for now as there is uncertainty with changing requirements. That function relies on the `EXAMINER_ACTIONS` the api provides to render the buttons.

PR #612, opened 2025-03-13
added the url colon action

thorwolpert:
*Issue:*
Added the :action to the permit url
Did not mount the permit under v1


thorwolpert:
next steps, remove the decorators on the called permit actions, 
change the mount point
update the URL called in the testcases.

PR #613, opened 2025-03-13
add path

avni-work:
*Issue:*

- bcgov/entity/issues/

*Description of changes:*

### DIFF
--- a/.github/workflows/strr-ui-cd.yaml
+++ b/.github/workflows/strr-ui-cd.yaml
@@ -9,7 +9,6 @@ on:
       - release*
     paths:
       - "strr-web/**"
-      - "strr-base-web/**"
   workflow_dispatch:
     inputs:
       target:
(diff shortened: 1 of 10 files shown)


.github/workflows/strr-ui-ci.yaml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
     paths:
       - "strr-web/**"
-      - "strr-examiner-web/**"
```
dimak1, line 7: The `base` is not part of it app

.github/workflows/strr-ui-cd.yaml
```diff
@@ -9,6 +9,7 @@ on:
       - release*
     paths:
       - "strr-web/**"
+      - "strr-base-web/**"
```
dimak1, line 12: This is also not needed here, this is legacy and not using Nuxt layers

PR #619, opened 2025-03-17
API - Registration Number Filter

shaangill025:
*Issue:*

- https://github.com/bcgov/entity/issues/26261

*Description of changes:*
- [x] Search by registration or application number
- [x] Application & registration status filter   


shaangill025:
@kris-daxiom Added test that covers:
- Creating an application, then searching on applicationNumber
- approving the application, then searching on registrationNumber
- searching application by `ACTIVE` status
- searching on an invalid recordNumber and returning no applications as expected

PR #628, opened 2025-03-21
Fix scroll in To Do for Pending NOC

dimak1:
*Issue:*

- bcgov/entity/issues/26622

*Description of changes:*
- Rework scroll to Supporting Info section in To Do for Pending NOC

### DIFF
--- a/strr-base-web/app/utils/todoItems.ts
+++ b/strr-base-web/app/utils/todoItems.ts
@@ -67,8 +67,10 @@ export const getTodoApplication = (
       newLine: '<br/>',
       boldStart: '<strong>',
       boldEnd: '</strong>',
-      linkStart: "<a href='#summary-supporting-info' class='text-blue-500 underline'>",
-      linkEnd: '</a>'
+      linkStart: "<button type='button'" +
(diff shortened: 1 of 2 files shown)


strr-base-web/app/utils/todoItems.ts
```diff
@@ -67,8 +67,10 @@ export const getTodoApplication = (
       newLine: '<br/>',
       boldStart: '<strong>',
       boldEnd: '</strong>',
-      linkStart: "<a href='#summary-supporting-info' class='text-blue-500 underline'>",
-      linkEnd: '</a>'
+      linkStart: "<button type='button'" +
```
dimak1, line 70: Used button for accessibility (so user can tab into it)

shaangill025: So breadcrumb issue was only triggered if tabbing into Todo?

dimak1: No, originally anchor links were modifying the url (by adding the id of the element to scroll). Not sure why breadcrumb was getting updated on url change... 
I just did a different solution, which is a smooth scroll without url update / anchor link.

shaangill025 (approved):
LGTM

PR #659, opened 2025-04-01
Examiner UI - Default Statuses

shaangill025:
*Issue:*

- https://github.com/bcgov/entity/issues/26657

*Description of changes:*
Follow up from https://github.com/bcgov/STRR/pull/658

### DIFF
--- a/strr-examiner-web/app/stores/examiner.ts
+++ b/strr-examiner-web/app/stores/examiner.ts
@@ -31,6 +31,26 @@ export const useExaminerStore = defineStore('strr/examiner-store', () => {
   const sendNocSchema = computed(() => z.object({
     content: z.string().min(1, { message: t('validation.nocContent') })
   }))
+  const defaultApplicationStatuses = [
+    ApplicationStatus.FULL_REVIEW,
+    ApplicationStatus.PROVISIONAL_REVIEW,
+    ApplicationStatus.PAYMENT_DUE,
+    ApplicationStatus.PROVISIONAL,
+    ApplicationStatus.PAID,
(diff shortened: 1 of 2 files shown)


dimak1:
@shaangill025 this PR does not cover all the task items in the ticket, right?

strr-examiner-web/app/stores/examiner.ts
```diff
@@ -31,6 +31,26 @@ export const useExaminerStore = defineStore('strr/examiner-store', () => {
   const sendNocSchema = computed(() => z.object({
     content: z.string().min(1, { message: t('validation.nocContent') })
   }))
+  const defaultApplicationStatuses = [
```
dimak1, line 34: What are these default statuses for?

shaangill025:
> @shaangill025 this PR does not cover all the task items in the ticket, right?

Only the draft applications issue, all else is already done and is in main.

bcregistry-sre:
Temporary Url for review: https://strr-examiner-dev--pr-659-rehwb0qd.web.app

strr-examiner-web/app/stores/examiner.ts
shaangill025: If no status is explicitly selected then the applications of these statuses will be shown by default.
https://github.com/bcgov/STRR/pull/658

dimak1: Should we just define which statuses not to show by default?

shaangill025: The API filter works based on inclusion, so explicit statuses need to be included when calling the endpoint.

dimak1 (approved):

PR #736, opened 2025-05-15
API changes - Set Aside Application Refusal Decision

kris-daxiom:
*Issue:*

- bcgov/entity/issues/27287

*Description of changes:*


strr-api/src/strr_api/services/application_service.py
```diff
@@ -398,3 +399,20 @@ def update_host_unit_address(application: Application, unit_address: dict, user:
             visible_to_applicant=True,
         )
         return application
+
+    @staticmethod
+    def set_aside_decision(application: Application, set_aside_request: dict, user: User) -> Application:
```
dimak1, line 404: Should we check if application is in Refuse state before setting aside? Or that check would happen in UI?

PR #847, opened 2025-08-28
Update API payload for Renewals

dimak1:
*Issue:*

- bcgov/entity/issues/30415

*Description of changes:*
- Update API payload for Renewals


strr-host-pm-web/app/stores/hostApplication.ts
```diff
@@ -12,6 +12,7 @@ export const useHostApplicationStore = defineStore('host/application', () => {
     agreedToRentalAct: z.boolean().refine(val => val, { message: t('validation.required') }),
     agreedToSubmit: z.boolean().refine(val => val, { message: t('validation.required') })
   })
+  const { isRenewal, registrationId } = useRouterParams()
```
dimak1, line 15: As mentioned in the chat, I don't like the idea of using query params to get renewal status or reg id. For time being and testing, we can stick with it, but I will work on a new solution in a new PR.

shaangill025 (approved):

PR #960, opened 2025-12-15
API - Registration Full Text Search & User Search Endpoints

shaangill025:
*Issue:*

- https://github.com/bcgov/STRR/issues/1233

*Description of changes:*
image

### DIFF
--- a/strr-api/src/strr_api/models/dataclass.py
+++ b/strr-api/src/strr_api/models/dataclass.py
@@ -32,7 +32,7 @@
 """ Data Class
 """
 from dataclasses import dataclass
-from typing import List, Optional
+from typing import List
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


strr-api/src/strr_api/models/dataclass.py
```diff
@@ -32,7 +32,7 @@
 """ Data Class
 """
 from dataclasses import dataclass
-from typing import List
+from typing import List, Optional
```
thorwolpert: Optional isn't used anymore, pref is to use |

JazzarKarim:
There's a unit test that's failing though, the CI is failing.

shaangill025:
> There's a unit test that's failing though, the CI is failing.

As discussed, the unit tests in CI are failing due to a connection pool issue (captured in #31487). They are all passing locally.

strr-api/src/strr_api/models/dataclass.py
shaangill025: Done

PR #1016, opened 2026-02-03
feat: pass document dates in applications & registrations responses

JazzarKarim:
*Issue:*

- bcgov/entity/issues/32298

*Description of changes:*
Registrations endpoint, before:
before reg endpoint

After:
after reg endpoint

Applications endpoint, before:
image

After:
image

The fix for registrations was very simple, just grab the created date from the documents table if addedOn is not found.

The fix for applications was trickier though as we don't store documents in the documents table if there is no registration yet. I created a function that grabs the date from GCP storage. 


dimak1 (approved):

thorwolpert (approved):
nice addition!

PR #1027, opened 2026-02-09
feat(services): add initial cut of InteractionService to start on Q and RenewalJob changes

thorwolpert:
*Issue:* bcgov/entity#32393

*Description of changes:*
- add initial cut of InteractionService to start on Q and RenewalJob changes
- add fixtures to support the new service
- add pytest-check to support multiple test and reporting at end of entire test
- adding overload to support pyright checking


thorwolpert:
I totally disagree with SC on this, which I think is more pythonic than a large langy if or match block. Short and easy to read, imho:
```python
        event_type = (
            Events.EventType.REGISTRATION
            if registration_id
            else Events.EventType.APPLICATION
            if application_id
            else Events.EventType.USER
        )
```

PR #1030, opened 2026-02-10
fix: properly store the initiatedByPropertyManager in registrations

JazzarKarim:
*Issue:*

- bcgov/entity/issues/32154

*Description of changes:*
- Look in the registrations_json, if the initiatedByPropertyManager is found, store it in the registration under property manager


thorwolpert:
You should use | None instead of Optional[] going forward.

dimak1:
@JazzarKarim good news if it fixes it. I also found a workaround for PMs during the renewals, so we should be good for now. It is in the ticket.

PR #1037, opened 2026-02-12
feat: updated conditions filter logic to accurately filter results

JazzarKarim:
*Issue:*

- bcgov/entity/issues/32371

*Description of changes:*
- PR only:
image

- BL only:
image

- PR + BL:
image

- I tested my changes in the registrations table as well, they work.

- I also compared results with what we already had in DEV, the numbers check out:
When I select PR in DEV, I get 543 results.
In here with my new changes:
  - When I select PR only, I get 274 results
  - When I select PR + BL, I get 269 results
274 + 269 = 543 which makes sense. Just some fun math here to verify results 😁 

- Did some refactoring to the code to fix the Cognitive Complexity issue reported by SC:
image
 

### Ignore the above, new requirements have been set:
• Selecting BL should return only records with value 'BL' and 'BL + PR'
• Selecting PR should return only records with value 'PR' and 'BL + PR'
• Selecting BL, PR should return all records with values 'BL', 'PR', and 'BL + PR'

Implemented the above

### DIFF
--- a/strr-api/src/strr_api/models/rental.py
+++ b/strr-api/src/strr_api/models/rental.py
@@ -249,6 +249,104 @@ class Registration(Versioned, BaseModel):
         )
 
     @classmethod
+    def _handle_bl_pr_requirement(cls, req: str, requirement: list[str], host_conditions: list):
+        """If req is BL+PR or PR (when both selected), handle and return True; else False."""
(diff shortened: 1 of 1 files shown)


dimak1:
@JazzarKarim we going to hold off the merge until we the release right?

JazzarKarim:
> @JazzarKarim we going to hold off the merge until we the release right?

Yep, I'm not merging until after release

PR #1055, opened 2026-03-02
feat: updates to Alembic and AuthService to support running without Flask.

thorwolpert:
*Issue:* bcgov/STRR#1175

*Description of changes:*
- Change env.py for Alembic to allow running without Flask instantiated
- Change AuthService - getToken(cls) to allow invocation without Flask being instantiated.
- Update tests to match changes to in the root conftest

### DIFF
--- a/strr-api/poetry.lock
+++ b/strr-api/poetry.lock
@@ -844,6 +844,27 @@ files = [
 ]
 
 [[package]]
+name = "dnspython"
+version = "2.8.0"
+description = "DNS toolkit"
+optional = false
+python-versions = ">=3.10"
+groups = ["main"]
+files = [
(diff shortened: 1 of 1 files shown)


strr-api/pyproject.toml
```diff
@@ -26,7 +26,7 @@ pytest-env = "^1.1.3"
 coloredlogs = "^15.0.1"
 flask-httpauth = "^4.8.0"
 flasgger = "^0.9.7.1"
-pydantic = "^2.7.1"
+pydantic = {extras = ["email"], version = "^2.12.5"}
```
JazzarKarim: Should we update the API version here on line 3? I think if you rebase the version is now 0.3.5. I think it's a good idea for a minor patch

JazzarKarim (approved):
Aside from that one small comment, looks great! 👍

PR #1065, opened 2026-03-04
chore: updated the dependencies of auto & provisional approval jobs

JazzarKarim:
*Issue:*

- bcgov/STRR#1167

*Description of changes:*
After doing some investigation, i found out the jobs are using a very old version of the API (0.1.5):
image

I got this from getting the image tag (commit SHA) from the job in GCP (DEV).

This PR updates the jobs to use the latest version which is 0.3.7

This is what I wanted the most:
image


thorwolpert (approved):
Great!

PR #1067, opened 2026-03-05
fix: fixed newest api code that was causing jobs to fail

JazzarKarim:
*Issue:*

- bcgov/STRR#1154

*Description of changes:*
- This was causing the jobs to fail!

After trying and trying, I managed to run the job locally!
I made my API change, installed locally:
```strr-api = {path = "../../strr-api", develop = true}```

Everything I had to do:
- Made some fixes locally to the devcontainer.json & dockerfile
- Opened a dev container
- Installed everything
- Made fixes to be able to connect to `host.docker.internal`
- Port-forwarded the dev db
- Filled in my .env file and point to dev

Before (which is what we see in the logs in GCP):
image

Now 😎 :
image

Job now runs successfully!

### DIFF
--- a/strr-api/pyproject.toml
+++ b/strr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "strr-api"
-version = "0.3.7"
+version = "0.3.8"
 description = ""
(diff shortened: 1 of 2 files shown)


thorwolpert (approved):
LGTM

PR #1107, opened 2026-03-18
fix(ui): fixed strata hotel unit type not auto filled in hosts

JazzarKarim:
*Issue:*

- bcgov/STRR#1145

*Description of changes:*
When completing a renewal in hosts that has a strata hotel exemption, the dropdown is disabled and users can't proceed anymore. This fixes so that the watcher is actually fired on initial load and not only on change which is the case for renewals. 


bcregistry-sre:
Temporary Url for review: https://strr-hosts-dev--pr-1107-cn33xgjy.web.app

JazzarKarim:
@dimak1 Dima, I tested all flows, this seems to be OK. I didn't see an issue with having this in!

Issue #1095, opened 2025-03-25
Bug: Odd behaviour when tabbing through the form and the CRA field


ser-popov:
Repro Steps:
1. Login as a host and start a new app
2. Go to Step 2
3. Choose to [Add an Individual] and 'I am adding my own info'
4. Enter a name and click on the [Tab] button on your keyboard
5. Fill in all required fields and tab next until you get to the CRA input
6. Use your mouse to check the exempt box

Actual:
See screenshot
The CRA input is red with an error text
But the exempt box is checked and the system will allow you to proceed
Just looks goofy

Expected:
Once the box is checked, no validations are performed on that input

<img width="1515" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOWRwQ1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--63607caee92586d756b400d1e0b90651b2aba865/image.png" alt="image.png" />

PR #1108, opened 2026-03-18
fix(host-pm-web): clear CRA tax number field error when exempt is checked

jimmypalelil:
*Issue:*

- bcgov/strr#1095

*Description of changes:*

- When "This individual does not have a CRA Tax Number" is checked, clear the owner form's `taxNumber` field error state via `ownerFormRef.clear('taxNumber')`, so the CRA input no longer stays in error state.


dimak1:
@jimmypalelil pls update the linked ticket in the description. I think it should be from `str` board and not from `entity`

PR #1456, opened 2026-03-27
feat(ui): hide action buttons for applications with reg number

dimak1:
*Issue:*

- bcgov/strr/issues/1436

*Description of changes:*
- Hide action buttons for applications with reg number


thorwolpert (approved):
looks like it covers it all

PR #1522, opened 2026-04-10
fix(ui): small fix to update payload during address requirements check

dimak1:
*Issue:*

- bcgov/strr/issues/1508

*Description of changes:*
- Update the addressLineTwo to be an empty string instead of null (in certain cases) to allow for successful address check


Jacky-Pham (commented):
Looks good to me

Jacky-Pham (approved):

PR #1532, opened 2026-04-15
test(strr-api): add pytests for api service

jimmypalelil:
*Issue:*

- bcgov/STRR/issues/1124

*Description of changes:*
unit tests for strr-api service

### how to test
```sh

cd strr-api
make test

```

#### Expected result

image

---


strr-api/pyproject.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 packages = [{include = "strr_api", from = "src"}]
 
 [tool.poetry.dependencies]
-python = "^3.11"
+python = "^3.12"
```
JazzarKarim: Is this something we need to do? Just need to be careful that this doesn't break the API from running.

Were you able to run the API normally with this version of python installed?

jimmypalelil: Actually it was supposed to be 3.12.2 to be consistent with the other serivces that depend on strr-test-utils. it was never required to be upgraded until this dependency. thanks for the highlight tho

I have tested the api service itself by running it locally.

JazzarKarim: Sounds good, thanks!

JazzarKarim (approved):
Nice, LG 👍

PR #1565, opened 2026-04-22
add Nanoid as a dependancy

Jacky-Pham:
*Issue:*

In relation to the document upload fix. things won't build unless we have this dependancy

- bcgov/entity/issues/

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the BC Registry and Digital Services BSD 3-Clause License

PR #1568, opened 2026-04-22
fix(strr-api): add nanoid as a runtime dependency

Jacky-Pham:
## Summary
- move `nanoid` from the test dependency group to runtime dependencies
- refresh `poetry.lock` to match the dependency move
- keep the change isolated to the API dependency files only

## Why
`nanoid` is imported by runtime application code in `strr-api/src/strr_api/models/application.py`, so keeping it in the test group can cause the deployed service to fail at startup with `ModuleNotFoundError`.

## Notes
This is a clean follow-up to #1565 with only the `nanoid` dependency change.


strr-api/pyproject.toml
```diff
@@ -32,6 +32,7 @@ google-cloud-storage = "2.14.0"
 geoalchemy2 = "^0.15.1"
 weasyprint = "^62.3"
 sqlalchemy-utils = "^0.41.2"
+nanoid = "^2.0.0"
```
JazzarKarim, line 35: @jimmypalelil had a question on this on the other PR: https://github.com/bcgov/STRR/pull/1565#discussion_r3125506462

jimmypalelil (approved):

JazzarKarim (approved):

PR #1587, opened 2026-04-27
fix(api): 1581 fix performance degradation issue with filtering in prod

JazzarKarim:
*Issue:*

- bcgov/strr/issues/1581

*Description of changes:*
I ran these changes locally:
BEFORE:

AFTER:

I tried running and testing via prod replica but i don't have write access:
image

### DIFF
--- a/strr-api/migrations/versions/20260427_1100_b1a4d5e7c92f_add_substatus_search_indexes.py
+++ b/strr-api/migrations/versions/20260427_1100_b1a4d5e7c92f_add_substatus_search_indexes.py
@@ -11,15 +11,15 @@ falling back to sequential scans on the ``registrations`` table.
 This migration adds the indexes required to turn those access paths into
 index scans:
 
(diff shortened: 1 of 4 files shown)


dimak1 (approved):

Jacky-Pham (commented):
I see some of the new migration includes non-ASCII em dashes on line 14. Can you replace these with regular hyphens. They're not functionally scary but for uniformity sake we should use regular hyphens.

The rental.py file also has this issue (line 482)

If it's whatever then lemme know i'll just approve cause functionally LGTM

JazzarKarim:
> I see some of the new migration includes non-ASCII em dashes on line 14. Can you replace these with regular hyphens. They're not functionally scary but for uniformity sake we should use regular hyphens.
> 
> The rental.py file also has this issue (line 482)
> 
> If it's whatever then lemme know i'll just approve cause functionally LGTM

Fixed, replaced with regular hyphens, thanks!

Jacky-Pham (approved):
LGTM

PR #1595, opened 2026-04-29
fix(api): 1392 send email to each recipient as separate requests

JazzarKarim:
*Issue:*

- bcgov/strr/issues/1547

*Description of changes:*
- Updated STRR email delivery so each recipient is sent as a separate notify-api request
  - Now, valid recipients still receive their email even if one or more recipients are invalid
- Added tests

### DIFF
--- a/queue_services/strr-email/src/strr_email/resources/email_listener.py
+++ b/queue_services/strr-email/src/strr_email/resources/email_listener.py
@@ -226,12 +226,20 @@ def worker():
         last_failure_status = None
         for recipient in recipients_list:
             single_email = {**email, "recipients": recipient}
-            resp = requests.post(
-                current_app.config["NOTIFY_SVC_URL"],
-                json=single_email,
-                headers=headers,
-                timeout=current_app.config["NOTIFY_API_TIMEOUT"],
-            )
+            try:
+                resp = requests.post(
+                    current_app.config["NOTIFY_SVC_URL"],
+                    json=single_email,
+                    headers=headers,
+                    timeout=current_app.config["NOTIFY_API_TIMEOUT"],
+                )
+            except Exception as err:  # noqa: BLE001
+                logger.info(f"Error sending recipient {recipient}: {str(err)}")
+                logger.error(
+                    f"Error posting email to notify-api for recipient {recipient}: {str(ce)}"
+                )
+                last_failure_status = HTTPStatus.BAD_REQUEST
+                continue
             if resp.status_code in [HTTPStatus.OK, HTTPStatus.ACCEPTED, HTTPStatus.CREATED]:
                 success_count += 1
             else:
@@ -243,12 +251,13 @@ def worker():
                 logger.error(
                     f"Error posting email to notify-api for recipient {recipient}: {str(ce)}"
                 )
-                last_failure_status = resp.status_code
+                # Return a 4xx from this worker for recipient-level notify failures.
+                last_failure_status = HTTPStatus.BAD_REQUEST
 
         if success_count == 0:
             return (
                 jsonify({"message": "Error posting email to notify-api."}),
-                last_failure_status or HTTPStatus.BAD_GATEWAY,
+                last_failure_status or HTTPStatus.BAD_REQUEST,
             )
 
     logger.info("completed ce (event_id=%s): %s", event_id, ce)
--- a/queue_services/strr-email/src/strr_email/resources/email_listener.py
+++ b/queue_services/strr-email/src/strr_email/resources/email_listener.py
@@ -208,26 +208,48 @@ def worker():
             return jsonify({"message": "Unable to send renewal reminder email."}), 400
 
     else:
-        token = AuthService.get_service_client_token()
-        resp = requests.post(
-            current_app.config["NOTIFY_SVC_URL"],
-            json=email,
-            headers={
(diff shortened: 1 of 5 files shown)


queue_services/strr-email/src/strr_email/resources/email_listener.py
```diff
@@ -194,21 +194,48 @@ def worker():
...
-        )
+        headers = {
+            "Content-Type": "application/json",
+            "Authorization": f"Bearer {token}",
+        }
+        success_count = 0
+        last_failure_status = None
+        for recipient in recipients_list:
+            single_email = {**email, "recipients": recipient}
+            resp = requests.post(
```
jimmypalelil: This api call could results in errors that could lead to 5xx errors. is that intended?

JazzarKarim: Actually Jimmy, good point. I've changed this to a return an HTTP 400 (Bad Request) if sending to particular recipient fails. I think that would be more accurate and makes more sense.

I've also added some logging to indicate which recipient it was.

strr-api/src/strr_api/services/interaction.py
```diff
@@ -120,7 +120,7 @@ def dispatch(
         interaction.registration_id = registration_id
         interaction.customer_id = customer_id
         interaction.user_id = user_id
-        interaction.notify_reference = notify_id
+        interaction.notify_reference = notify_json.get("ids") or notify_id
```
jimmypalelil: sanity check: we are modifying the shape of the `notify_reference`  here. is this clear of any downstream side effects?

queue_services/strr-email/src/strr_email/resources/email_listener.py
jimmypalelil: i am not sure about 400 as that would imply user error. I wouldnt do that unless we are absolutely sure that the error results from bad email (which would be a user error). what do you think?

JazzarKarim: I was thinking that the error does indeed stem from a user error right? Since the email was poorly formatted (input was something wrong), like @ba$.com which is invalid. The sending of email fails in this case. That's why I was thinking in this a 4xx makes sense? Thoughts?

jimmypalelil: The try catch logic will catch any excpetion and throw 4xx. I am not sure if thats the right behaviour

JazzarKarim: OK, good point, thanks. Fixed.

Now, if the error is from the recipient's email being badly formatted, it's a 400 error.
Otherwise, it's a 502 error (Bad Gateway).

This should make the error message not to be misleading.

strr-api/src/strr_api/services/interaction.py
JazzarKarim: I changed this to make it backwards compatible as a single notify ID again. That should be much safer. Thanks Jimmy.

jimmypalelil (approved):

strr-api/src/strr_api/services/interaction.py
```diff
@@ -169,11 +174,43 @@ def queued(
...
+            result = InteractionService._post_email_to_notify(single_email)
+            notify_id = result.get("id") if isinstance(result, dict) else None
+            if notify_id and not (isinstance(notify_id, (int, float)) and notify_id < 1):
+                success_ids.append(notify_id)
+                last_result = result
+
+        if not success_ids:
+            return {"id": -1}
+
+        combined_ids = ",".join(str(i) for i in success_ids)[:100]
```
dimak1, line 203: So if there are more recipients that fit in 100 chars, they will be dropped?

queue_services/strr-email/src/strr_email/resources/email_listener.py
```diff
@@ -199,36 +210,75 @@ def worker():
...
+            try:
+                resp = requests.post(
+                    current_app.config["NOTIFY_SVC_URL"],
+                    json=single_email,
+                    headers=headers,
+                    timeout=current_app.config["NOTIFY_API_TIMEOUT"],
+                )
+            except Exception as err:  # noqa: BLE001
+                logger.info(f"Error sending recipient {recipient}: {str(err)}")
+                logger.error(
```
dimak1, line 249: Do we need `info` logger above if we logging `error` here?
Same question on line 262 below.

strr-api/src/strr_api/services/interaction.py
JazzarKarim: Yes, only for that stored string though, not for sending. Every recipient will still get their own POST call.

queue_services/strr-email/src/strr_email/resources/email_listener.py
JazzarKarim: The first one logs the actual exception (timeout for example or some connection issue). The second one logs the cloud event object for this pub/sub message. TBH I can combine both, but felt it would be nice to make this separation. Let me know if you want me to combine them Dima!

dimak1: All good, thanks for clarifying.

dimak1 (approved):

Issue #1543, opened 2026-04-16
strr-api: add integration test suite (tests/integration, pytest marker, CI)


jimmypalelil:
## Summary

Introduce a **dedicated integration test suite** for **strr-api** that exercises the Flask app against a **real PostgreSQL** instance (aligned with existing `testcontainers` usage in `tests/conftest.py`) and validates **end-to-end HTTP behaviour** beyond what unit tests cover with heavy mocking.

Unit test setup and coverage is already up to date and was covered in this ticket: [#1124](https://github.com/bcgov/STRR/issues/1124)

## Goals

- Add a clear **`tests/integration/`** (or equivalent) layout with shared fixtures (app + DB lifecycle, migrations, auth helpers).
- Cover **high-value API flows** that benefit from DB + routing together
- Ensure **CI** runs integration tests on PRs.

## Acceptance criteria

- [ ] Integration test package exists under `strr-api/tests/` with documented setup in `README` or `tests/README.md`.
- [ ] CI workflow executes the integration suite (or documents why it is scheduled/nightly with a follow-up ticket).

jimmypalelil:
This work will include a series of PRs. Here's the first one: https://github.com/bcgov/STRR/pull/1607

PR #1608, opened 2026-05-05
test(strr-api): add integration harness and shared seed utilities

jimmypalelil:
*Issue:* #1543

## Summary
- add integration test harness wiring in `tests/conftest.py` and `tests/integration/conftest.py`, using shared reusable fixture plugins from `strr_test_utils`
- add shared integration helpers and deterministic seed utilities in `tests/integration/helpers.py`, `tests/integration/registration_seed.py`, and `tests/integration/application_seed.py`
- include supporting integration config updates in `strr-api/Makefile`, `strr-api/pyproject.toml`, and `tests/python-test-utils/src/strr_test_utils/parent_fixtures.py`
- all the ci trigger paths have been updated since there is dependency on fixtures which comes from this path. so its best to run these ci checks just to ensure nothings broken with the new changes.
---

### DIFF
--- a/.github/workflows/strr-api-ci.yaml
+++ b/.github/workflows/strr-api-ci.yaml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - "strr-api/**"
+      - "tests/python-test-utils/**"
   workflow_dispatch:
 
 defaults:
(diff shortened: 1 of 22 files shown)


.github/workflows/strr-api-ci.yaml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - "strr-api/**"
+      - "tests/python-test-utils/**"
```
jimmypalelil, line 7: all the ci trigger paths have been updated since there is dependency on fixtures which comes from this path. so its best to run these ci checks just to ensure nothings broken with the new changes.

JazzarKarim (approved):
LGTM

PR #1621, opened 2026-05-07
fix(ui): 1604 fix strata hotel renewal applications

JazzarKarim:
*Issue:*

- bcgov/strr/issues/1604

*Description of changes:*
cherry picked the hotfix to main

### DIFF
--- a/strr-examiner-web/package.json
+++ b/strr-examiner-web/package.json
@@ -2,7 +2,7 @@
   "name": "strr-examiner-web",
   "private": true,
   "type": "module",
-  "version": "0.2.41",
+  "version": "0.2.42",
   "engines": {
     "node": ">=24"
   },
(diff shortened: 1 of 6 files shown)


strr-examiner-web/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "strr-examiner-web",
   "private": true,
   "type": "module",
-  "version": "0.2.41",
+  "version": "0.2.42",
```
dimak1, line 5: So the actual hotfix version is lost?

JazzarKarim: Yep. That version there in the hotfix is kind of a temporary measure to make sure the version is there. I figured once we're back here to main, we'll go with our normal flow with a minor patch. After cherry-picking, I made the update manually myself.

dimak1 (approved):

Issue #1041, opened 2026-02-20
Monitoring


thorwolpert:
In order to be as proactive as possible, monitoring, alerting and reporting on key areas needs to be implemented. There is other infrastructure monitoring in place managed by the core SRE team, but the items below are the projects responsibility to manage.

DLQ (Dead Letter Queue) - Items that are logged here have failed to be applied and require human intervention:
- [ ] strr-emailer-queue - emails that were unable to be processed by notify
- [ ] strr-pay-queue  - payments that have failed to be against applications

Interactions - failures logged here are when a notification has failed to be delivered to the receipient.
- [ ] interaction_status is FAILED or (SENT and an excessive time has passed)

OlgaPotiagalova:
Payment status visibility is more critical for renewals than for new applications. Payments that remain in an incomplete state can lead to listings being removed, requiring hosts to manage cancellations. Could we introduce separate categories for renewals and applications in monitoring?

Can we differentiate between specific states to determine that the payment is abandoned vs delayed vs failed (explicit statuses)?

Jacky-Pham:
Monitoring dashboards are set up now.

Email triage:
Dev: https://console.cloud.google.com/monitoring/dashboards/builder/d9131b5b-870a-4590-a39e-01a497d3f4de?project=bcrbk9-dev
Test: https://console.cloud.google.com/monitoring/dashboards/builder/62c4bddf-4675-4b47-90b1-1b68f54bb533?project=bcrbk9-test

Documents:
Dev: https://console.cloud.google.com/monitoring/dashboards/builder/3958ddf9-896f-4a19-8ded-83548692088a?project=bcrbk9-dev
Test: https://console.cloud.google.com/monitoring/dashboards/builder/489773b7-15ae-40de-83eb-6bc7e7823847?project=bcrbk9-test

Payments:
Dev: https://console.cloud.google.com/monitoring/dashboards/builder/7cecae79-91a0-4243-b52a-bea70d313be0?project=bcrbk9-dev
Test: https://console.cloud.google.com/monitoring/dashboards/builder/ab9d5f42-a43e-4eaa-ba46-79a8763f2f1f?project=bcrbk9-test

Jacky-Pham:
Prod dashboards are set up now as well.

Email flow: https://console.cloud.google.com/monitoring/dashboards/builder/005a0a2a-1dc7-487b-9c2e-1a6ba730acfc?project=bcrbk9-prod
[...]

PR #1640, opened 2026-05-13
#1041 feat(observability): add failure monitoring signals

Jacky-Pham:
## Summary
- added local failure-monitoring signal commit to the PR branch
- included the local strata hotel active registration email support commit already added earlier
- retained the empty tracking commit for the dev pay queue monitoring work configured in GCP
- keeps docs/runbooks/observability.md out of the PR

## Notes
- f73d4823 was already represented in origin/main, so it was not duplicated
- 8af2fdb6 was cherry-picked as 35b88a56
- fde38a0a was cherry-picked and conflict-resolved as 6e69358e

## Validation
- python3 -m py_compile queue_services/strr-email/src/strr_email/resources/email_listener.py strr-api/src/strr_api/services/interaction.py strr-api/tests/unit/services/test_interaction_service.py queue_services/strr-email/tests/conftest.py

Refs #1041


queue_services/strr-email/src/strr_email/resources/email_listener.py
```diff
@@ -193,96 +182,34 @@ def worker():
...
+            "cloud_event_id=%s email_type=%s application_number=%s registration_number=%s "
+            "interaction_uuid=%s interaction_key=%s error=%s",
+            getattr(ce, "id", None),
+            email_info.email_type,
+            email_info.application_number,
+            email_info.registration_number,
+            email_info.interaction_uuid,
+            email_info.interaction,
+            err,
+        )
```
jimmypalelil, line 213: The recipient list should be split in order to avoid the errors related to typos in email addresses

queue_services/strr-email/src/strr_email/resources/email_listener.py
```diff
@@ -193,96 +182,34 @@ def worker():
...
+            "strr.email.notify.failed Error posting email to notify-api "
+            "cloud_event_id=%s email_type=%s application_number=%s registration_number=%s "
+            "interaction_uuid=%s interaction_key=%s error=%s",
+            getattr(ce, "id", None),
+            email_info.email_type,
+            email_info.application_number,
+            email_info.registration_number,
+            email_info.interaction_uuid,
+            email_info.interaction,
+            err,
```
jimmypalelil, line 211: traceback was added to get the complete stack trace of the error. This helps in debugging

strr-api/src/strr_api/resources/documents.py
```diff
@@ -86,7 +86,16 @@ def upload_supporting_document():
...
         filename = secure_filename(file.filename)
-        document = DocumentService.upload_document(filename, file.content_type, file.read())
+        document = DocumentService.upload_document(
+            filename,
+            file.content_type,
+            file.read(),
+            metadata={
+                "upload_source": "documents",
+                "document_type": "SUPPORTING_DOCUMENT",
+                "upload_step": "standalone_supporting_document",
```
jimmypalelil, line 99: I think we should add user metadata here as well. Just having these three attributes wont provide much info.

The document type should specify which one from the UI was selected.

strr-api/src/strr_api/services/email_service.py
```diff
@@ -65,6 +65,43 @@
...
+    @staticmethod
+    def _publish_tracked_email(
+        *,
+        payload_data: dict,
+        application_id: int | None = None,
+        registration_id: int | None = None,
+        idempotency_key: str | None = None,
+    ):
+        """Create a queued interaction before publishing the email event."""
+        # TODO: fix circular import issue
```
jimmypalelil: TODO left here. if the circular dependency has been resolved then these imports can be moved to the top

jobs/interactions-update/src/interactions_update/job.py
```diff
@@ -26,7 +29,7 @@ def fetch_and_update(interaction_id, notify_url, headers, timeout):
     try:
         interaction = session.get(CustomerInteraction, interaction_id)
         if not interaction or not interaction.notify_reference:
-            return False
+            return "missing_reference"
```
jimmypalelil: QQ Whats the reason for changing the return type from boolean to random string values?

jobs/interactions-update/src/interactions_update/job.py
```diff
@@ -79,12 +82,31 @@ def run(max_workers=None):
     session = next(session_gen)
 
     try:
+        stale_sent_hours = int(os.getenv("STALE_SENT_HOURS", "24"))
+        stale_threshold = datetime.now(timezone.utc) - timedelta(hours=stale_sent_hours)
+        stale_stmt = select(CustomerInteraction.id).where(
```
jimmypalelil: there is another select statement below querying the same table... would it benefit to merge into one query and do comparison in here instead?

strr-api/src/strr_api/resources/application.py
```diff
@@ -781,7 +781,18 @@ def upload_registration_supporting_document(application_number):
...
-        document = DocumentService.upload_document(filename, file.content_type, file.read())
+        document = DocumentService.upload_document(
+            filename,
+            file.content_type,
+            file.read(),
+            metadata={
+                "upload_source": "application",
+                "entity_id": application.id,
+                "application_number": application.application_number,
+                "document_type": "SUPPORTING_DOCUMENT",
```
jimmypalelil: same as the other comment... is there any value in hardcoding SUPPORTING_DOCUMENTS. This GCS bucket is meant for supporting documents and we already know that so its better to have the type from the UI selection by the user.

jimmypalelil:
@Jacky-Pham I think you have to update the poetry.lock files as well for both job and queue service since they depend on strrr-api. Without this, the latest changes wont be picked up in their build jobs

jobs/interactions-update/src/interactions_update/job.py
Jacky-Pham: I updated this so the job is clearer about what happened. Instead of returning False, it now returns a named result for “missing reference,” which makes the later logging/counting easier to understand.

strr-api/src/strr_api/resources/application.py
Jacky-Pham: Okay I updated it so the upload now reads documentType from the form data and stores that in the metadata instead of hardcoding SUPPORTING_DOCUMENTS. I left OTHERS as the fallback if the request does not include a type.

strr-api/src/strr_api/services/email_service.py
Jacky-Pham: TODO was removed and the imports were moved to the top of email_service.py. I also updated the interaction service import path to avoid the circular dependency

Jacky-Pham:
Hey @JazzarKarim, could you take a look at #1640 when you have a chance? It’s mergeable, checks are green, and Jimmy’s lock-file feedback has been handled. I’m trying to close out #1041 and unblock the document upload observability work.

jobs/interactions-update/src/interactions_update/job.py
jimmypalelil: But what thats the point of these changes? If you wanted to use an enum then wouldnt it be better to use `InteractionStatus` instead?

jobs/interactions-update/src/interactions_update/job.py
```diff
@@ -107,6 +140,22 @@ def run(max_workers=None):
                     )
                     for i_id in interaction_ids
                 ]
-                concurrent.futures.wait(futures)
+                for future in concurrent.futures.as_completed(futures):
+                    results.append(future.result())
+
+        failed_count = results.count(InteractionStatus.FAILED)
```
jimmypalelil, line 148: should we count the missing refrences here as well?

jimmypalelil:
@Jacky-Pham thansk for making changes to the PR. could you please reply to the comments I have made above. Its hard to figure out if all of them are addressed just by looking at the refreshed code.

queue_services/strr-email/src/strr_email/resources/email_listener.py
```diff
@@ -193,96 +183,35 @@ def worker():
...
-            failure_status = (
-                HTTPStatus.BAD_GATEWAY if has_non_format_failure else HTTPStatus.BAD_REQUEST
-            )
-            return (
-                jsonify({"message": "Error posting email to notify-api."}),
-                failure_status,
-            )
-
-    logger.info("completed ce (event_id=%s): %s", event_id, ce)
-    return {}, HTTPStatus.OK
```
JazzarKarim, line 285: Question here, where is this implemented now? The sending an email to multiple recipients and if one of them fails because of bad email format, the others are processed normally

Jacky-Pham: Yep, this is implemented in `InteractionService` rather than in the email worker. The worker still builds the comma-separated recipient string and passes it through `InteractionService.dispatch(...)` the split/send behavior is in `strr-api/src/strr_api/services/interaction.py` and `_send_email_to_notify_service`.

That method splits the comma-separated `recipients`, sends one Notify request per recipient, collects successful Notify ids, and only returns failure if all recipient sends fail. The focused coverage is in `strr-api/tests/unit/services/test_interaction_service.py`:
- `test_dispatch_email_splits_recipients`
- `test_dispatch_email_partial_recipient_failure`
- `test_dispatch_email_all_recipients_fail`

If that answers the question, could you approve this one when you get a chance? Checks are green now and this is the last blocker before I can merge/close #1041.

JazzarKarim: Nice, OK, thanks!

JazzarKarim (commented):
LGTM. I see a comment from Jimmy there, once everything is addressed, we can merge. Thanks!

jobs/interactions-update/src/interactions_update/job.py
Jacky-Pham: Yep, I folded this into the same SENT interaction query. It now selects the id plus derived `is_stale` and `missing_notify_reference` flags in one pass, then only sends Notify lookups for rows that have a reference.

jobs/interactions-update/src/interactions_update/job.py
Jacky-Pham: Good call. I removed the custom `UpdateResult` enum and now return the actual `InteractionStatus` only when Notify moves a record to a terminal status; unchanged/missing-reference rows return `None`.

jobs/interactions-update/src/interactions_update/job.py
Jacky-Pham: Added this. The completion log now includes `missing_reference_count`, and there is a focused test covering SENT interactions that are skipped because they do not have a Notify reference.

jobs/interactions-update/src/interactions_update/job.py
Jacky-Pham: Agreed. I dropped the custom enum here. The job now uses `InteractionStatus` for real status transitions and `None` for no status change.

queue_services/strr-email/src/strr_email/resources/email_listener.py
Jacky-Pham: This is handled now through `InteractionService.dispatch(...)`. The email worker passes the comma-separated recipients through, and `InteractionService._send_email_to_notify_service` splits them into individual Notify requests so one bad address does not block the others.

strr-api/src/strr_api/resources/documents.py
Jacky-Pham: Added richer metadata here: account id, selected `documentType`, upload step, and user token details. The standalone upload test now asserts those fields are passed to storage.

queue_services/strr-email/src/strr_email/resources/email_listener.py
Jacky-Pham: Kept the full traceback in the structured failure log so we still have stack detail when the worker fails posting/dispatching email.

strr-api/src/strr_api/services/email_service.py
Jacky-Pham: Cleaned this up. The circular import issue is resolved and the interaction imports are now at module scope.

strr-api/src/strr_api/resources/application.py
Jacky-Pham: Updated this to use the selected `documentType` from the upload form, defaulting to `OTHERS` only when the client does not provide one. The application upload metadata test covers that now.

jimmypalelil (approved):
Looks good now.. thanks for the changes 👍

JazzarKarim (approved):
LG, great stuff Jacky!

PR #1650, opened 2026-05-20
fix(api): fixed review/review renew substatus filter bug

JazzarKarim:
*Issue:*

- bcgov/strr/issues/1639

*Description of changes:*
There was a gap in my changes. Needed API change as well.

I was testing in TEST, created some data manually, here's what i found:
image

I was working with registration: H329198221
As you can see, the sub-status now shows approved, but the Review sub-status still showed it. In order to fix that, the API now checks the latest app decider as well per the legacy flow.

Now, while running locally, it doesn't show it anymore:
image

### DIFF
--- a/strr-api/pyproject.toml
+++ b/strr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "strr-api"
-version = "0.3.17"
+version = "0.3.17c"
 description = ""
 authors = ["thorwolpert <thor@wolpert.ca>"]
 license = "BSD 3-Clause"
--- a/strr-api/src/strr_api/models/rental.py
+++ b/strr-api/src/strr_api/models/rental.py
@@ -182,6 +182,22 @@ class Registration(Versioned, BaseModel):
         return sub_status_conditions
 
     @classmethod
+    def _latest_application_field_subquery(cls, application_field):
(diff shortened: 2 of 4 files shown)


strr-api/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "strr-api"
-version = "0.3.17"
+version = "0.3.17c"
```
JazzarKarim, line 3: matches package.json in examiner now

strr-api/src/strr_api/models/rental.py
```diff
@@ -473,14 +473,25 @@ def _approval_method_condition(cls, approval_methods: list[str], examiner_review
             .limit(1)
             .scalar_subquery()
         )
+        latest_app_decider_subq = (
```
dimak1: Can this and `latest_app_status_subq` above be combined, as they are hitting the same row?

dimak1 (approved):
LGTM with one comment.

strr-api/src/strr_api/models/rental.py
JazzarKarim: Good catch Dima. I refactored to decrease the duplicated code. Thanks!

PR #1657, opened 2026-05-22
fix(api): 1639 fixed approved substatus

JazzarKarim:
*Issue:*

- bcgov/strr/issues/1639

*Description of changes:*
align registration sub-status filtering with dashboard logic by treating latest application level decider as reviewed for provisional records, while keeping auto/full approvals included regardless of decider

### DIFF
--- a/strr-api/pyproject.toml
+++ b/strr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "strr-api"
-version = "0.3.18d"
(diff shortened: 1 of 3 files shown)


dimak1 (approved):
LG. Any idea if this might slow down the system when executed at scale?

JazzarKarim:
> LG. Any idea if this might slow down the system when executed at scale?

Good point. No, this should work the same speed/efficiency as what we had before.

PR #1669, opened 2026-05-27
fix: add pg8000 graceful shutdown on Cloud Run scale-down

panish16:
## Summary

- Registers a SQLAlchemy `connect` event listener on app startup that wraps each pg8000 connection's `close()` to suppress `pg8000.exceptions.InterfaceError` during Cloud Run scale-down
- Fixes intermittent `pg8000.exceptions.InterfaceError: network error` logged when Cloud Run terminates idle instances and SQLAlchemy's connection pool teardown tries to close already-severed sockets
- `strr-email` and `strr-pay` connect via Unix socket (no Cloud SQL Connector dependency), so the listener is implemented inline using the same logic

## Services fixed

- `queue_services/strr-email/src/strr_email/__init__.py`
- `queue_services/strr-pay/src/strr_pay/__init__.py`

## Test plan

- [ ] Confirm no new `pg8000.exceptions.InterfaceError` entries in GCP Log Explorer for `bcrbk9-prod` project after deployment
- [ ] Existing unit and integration test suites pass


Jacky-Pham (approved):
LGTM

dimak1 (approved):

queue_services/strr-email/poetry.lock
```diff
@@ -1,4 +1,186 @@
-# This file is automatically @generated by Poetry 2.3.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.3.2 and should not be changed by hand.
```
jimmypalelil: poetry version is mismatched. is that intended?

queue_services/strr-pay/poetry.lock
```diff
@@ -1,4 +1,186 @@
-# This file is automatically @generated by Poetry 2.3.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.3.2 and should not be changed by hand.
```
jimmypalelil: same comment here as above about the poetry version

.sonarcloud.properties
```diff
@@ -3,7 +3,7 @@ sonar.projectVersion=Autoscan
 
 # Path to sources
 #sonar.sources=strr-api/src/**/*
-sonar.exclusions=strr-api/migrations/**/*,strr-api/devops/**/*,strr-api/tests/**/*,testing/**/*
+sonar.exclusions=strr-api/migrations/**/*,strr-api/devops/**/*,strr-api/tests/**/*,testing/**/*,queue_services/**/tests/**/*
```
jimmypalelil, line 6: whats the reason for adding this?

jimmypalelil:
hey @panish16 thanks for integrating the `cloud-sql-connector` lib. Its only applied to couple python proejcts. Should it not be applied to all that depend on the DB?

queue_services/strr-email/tests/integration/test_queue_email.py
```diff
@@ -379,3 +379,121 @@ def notify_callback(request):
     assert stored.registration_id == registration.id
     assert stored.status == InteractionStatus.SENT
     assert stored.meta_data["email_type"] == "STRATA_HOTEL_REGISTRATION_ACTIVE"
+
+
+def test_email_no_cloud_event_data(client, queue_envelope, simple_cloud_event):
```
jimmypalelil: what are these unit & integration tests added for in this project?

wasnt it already at 100% coverage?

queue_services/strr-email/poetry.lock
panish16: Not intentional — the lockfile was regenerated with my local Poetry 2.3.2 when I added the `cloud-sql-connector` dependency. The 2.3.4 header was from the previous lockfile. The lock content (packages and hashes) is the same; only the generator version comment in line 1 differs. I can upgrade Poetry locally to 2.3.4 and regenerate if you'd prefer consistency.

queue_services/strr-pay/poetry.lock
panish16: Same as above — same unintentional regeneration with local Poetry 2.3.2. Content is correct.

.sonarcloud.properties
panish16: SonarCloud was picking up the new test files added under `queue_services/` as source files and reporting 0% coverage on them (the main SonarCloud scan doesn't run the queue_services test suites separately). Without the exclusion, the quality gate would flag those test files as unmeasured source. This follows the same pattern as the existing `testing/**/*` exclusion already in the file.

queue_services/strr-email/tests/integration/test_queue_email.py
panish16: Removing these tests — on closer inspection the session-scoped `app` fixture already calls `create_app(TestConfig)` once for the entire test suite, so coverage stays at 100% without them. The tests were also testing pre-existing `email_listener.py` utilities unrelated to this fix, so they're out of scope for this PR.

panish16:
@jimmypalelil Good question! To clarify why the fix is scoped to just `strr-email` and `strr-pay`:

- **`strr-api`** — uses **psycopg2** in production (the base `Config` class sets `postgresql+psycopg2://...`). Only the `Migration` config class uses pg8000, and migrations are short-lived (run to completion, no idle pool). So `strr-api` is not affected by the Cloud Run idle scale-down / pg8000 `InterfaceError` issue.
- **`jobs/*`** — batch jobs that run to completion and exit cleanly. Cloud Run terminates them normally, not via idle scale-down, so the connection pool teardown issue does not apply.
- **`queue_services/batch-permit-validator`** — has pg8000 in its deps transitively but does not initialize a SQLAlchemy engine in `create_app()`; no pool to attach the listener to.

`strr-email` and `strr-pay` are the only long-running Cloud Run services that hold a pg8000 connection pool while idling between queue messages, which is exactly where the Cloud Run scale-down kills the socket before SQLAlchemy can close it cleanly.

queue_services/strr-email/tests/integration/test_queue_email.py
jimmypalelil: Have you tried running `make test` in folder `queue_services/strr-email` without the unit & integration tests you added? its still at 100%. not sure why you are getting less coverage

jimmypalelil (approved):

PR #1671, opened 2026-05-27
chore(api): transfer Alembic object ownership

Jacky-Pham:
## Summary
This is an SRE initiative that came from the security team around Cloud SQL IAM/service account ownership.

I lined this up with Andriy's Alembic IAM migration flow from the other apps. The important bit is the same pattern used in PPR/Search: read DATABASE_OWNER_ROLE, escape quotes, run SET ROLE before Alembic runs, then commit so migrations create objects as that DB role.

So the short explanation is: I basically copied the migration ownership approach Andriy used there and adapted it to STRR's standalone env.py.

## What changed
- added DATABASE_OWNER_ROLE support in STRR Alembic env.py
- runs SET ROLE before online migrations when DATABASE_OWNER_ROLE is set
- escapes double quotes in the configured role name before building the SET ROLE statement
- documents that the migration user must be able to SET ROLE and the owner role needs schema create privileges
- added an integration test that runs Alembic against disposable Postgres with an IAM-style role name and verifies object/type ownership

## Testing
- python3 -m py_compile strr-api/migrations/env.py strr-api/tests/integration/test_alembic_ownership.py
- poetry run bandit -r migrations/env.py
- poetry run pytest tests/integration/test_alembic_ownership.py --no-cov
- targeted coverage check for migrations/env.py: 83%

Note: I also reran the full integration suite. The Alembic test passed, but two existing NOC tests failed on nocSentDate being null. I reran those exact two tests and they still failed, so I’m treating that as unrelated to this migration-role change.


dimak1 (approved):
LGTM

PR #1757, opened 2026-06-30
fix(ui): footer cta buttons layout

dimak1:
*Issue:*

- Recent changes to the button control layout broke the Examiner Details page, where the buttons would have empty space on the right
- 
Screenshot 2026-06-30 at 10 52 56

*Description of changes:*
- fix layout using css only, and revert previous change to remove the empty div block

### DIFF
--- a/strr-base-web/.eslintrc
+++ b/strr-base-web/.eslintrc
@@ -24,7 +24,8 @@
         "connect-date-picker",
         "connect-date-picker__err",
         "prose-bcGov",
-        "max-w-bcGovInput"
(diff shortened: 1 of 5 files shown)


bcregistry-sre:
Temporary Url for review: https://strr-hosts-dev--pr-1757-j1cu99wd.web.app

jimmypalelil (approved):
approving now since there is a bug thats merged in but long term fix needed

dimak1:
This could be treated as not only fix but also as re-work of the original requirement.

PR #77, opened 2024-07-30
Add 'max-width' for All Pages

dimak1:
*Issue:*
- bcgov/entity/issues/22484

*Description of changes:*
- Add max-width layout and other small fixes

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the namex license (Apache 2.0).

PR #1763, opened 2026-07-08
Move strr-api deployed DB access to Cloud SQL IAM auth

Jacky-Pham:
## Summary

This moves deployed `strr-api` database access from password/socket DB auth to Cloud SQL IAM database authentication through the Cloud SQL Auth Proxy.

Local Docker and pytest still use the existing password-based Postgres config. Normal local dev is unaffected as long as local env stays on `POD_NAMESPACE=local` or leaves `DEPLOYMENT_ENV` unset.

## Changes

- Removed `cloud-sql-python-connector[pg8000]` and kept the normal `psycopg2` SQLAlchemy path.
- Updated deployed GCP `strr-api` config to build a passwordless `postgresql+psycopg2://` URI for the proxy endpoint.
- Runtime API uses `DATABASE_USERNAME`; migration mode uses `DATABASE_MIGRATION_USERNAME` when `DEPLOYMENT_ENV=migration`.
- Deployed GCP env mapping now sets `DATABASE_HOST=127.0.0.1` and `DATABASE_PORT=5432` for the proxy sidecar.
- Local/test DB config still uses the existing password-based path.
- Updated the IAM config tests and the vault mapping guard for the proxy approach.
- Kept Alembic ownership aligned with latest `main` through `DATABASE_OWNER_ROLE`.

## Deployment Handoff

I do not have access to update these 1Password values directly, so this still needs to be handled by someone with access.

Needed STRR GCP env mapping changes:

Remove deployed use of:

- `DATABASE_PASSWORD`
- `DATABASE_UNIX_SOCKET`
- `CLOUDSQL_INSTANCE_CONNECTION_NAME`
- `CLOUDSQL_IP_TYPE`

Add/keep:

- `DATABASE_NAME=strr-db`
- `DATABASE_HOST=127.0.0.1`
- `DATABASE_PORT=5432`
- `DATABASE_USERNAME`
- `DATABASE_MIGRATION_USERNAME`
- `DATABASE_OWNER_ROLE=strr`

For UAT, use the `bcrbk9-test` identity and Cloud SQL instance because UAT deploys into the test project.

The Cloud Run deployment layer also needs to run Cloud SQL Auth Proxy v2 with automatic IAM database authentication, listening on `127.0.0.1:5432`, using the existing `cloudsql-instances` value for the instance connection name:

```sh
cloud-sql-proxy --auto-iam-authn --port=5432 <INSTANCE_CONNECTION_NAME>
```


bcregistry-sre:
Temporary Url for review: https://strr-hosts-dev--pr-1763-ciu8zc70.web.app

strr-api/src/strr_api/config.py
```diff
@@ -45,11 +45,93 @@
...
+    db_name = os.getenv("DATABASE_NAME", "")
+    db_host = os.getenv("DATABASE_HOST", "")
+    db_port = int(os.getenv("DATABASE_PORT", "5432"))
+
+    if db_unix_socket := os.getenv("DATABASE_UNIX_SOCKET", None):
+        if driver == "pg8000":
+            return f"postgresql+pg8000://{db_user}:{db_password}@/{db_name}?unix_sock={db_unix_socket}/.s.PGSQL.5432"
+        return f"postgresql+psycopg2://{db_user}:{db_password}@/{db_name}?host={db_unix_socket}"
+
+    if driver == "pg8000":
```
jimmypalelil: why are switching to `pg8000` from `psycopg2`. lets use auth proxy from gcp for the iam auth.

ref: https://docs.cloud.google.com/sql/docs/postgres/sql-proxy

Jacky-Pham: I had thought about this but figured it would be a more complex change and more resistance. I can switch to proxy

Jacky-Pham: switched to proxy please review

strr-api/tests/unit/test_config_cloudsql_iam.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+    assert (
+        config_module.Production.SQLALCHEMY_DATABASE_URI == "postgresql+psycopg2://sa-api%40bcrbk9-dev.iam@/strr-db"
+        "?host=%2Fcloudsql%2Fbcrbk9-dev%3Anorthamerica-northeast1%3Astrr-db-dev"
+    )
+    assert config_module.Production.SQLALCHEMY_ENGINE_OPTIONS == {}
+
+
+def test_deployed_config_requires_cloudsql_proxy_iam_env(monkeypatch):
```
jimmypalelil: this test case is failing for me in local. do i have configure the .env in a certain way?

Jacky-Pham: sorry you shouldn’t need to configure your .env for this test. I reproduced the failure with a normal local .env and the config reload was loading those values after the test cleared the environment, so the RuntimeError was not raised. I updated the test helper to disable dotenv loading

strr-api/src/strr_api/config.py
```diff
@@ -45,11 +45,79 @@
...
+
+
+def _database_settings() -> tuple[str, dict]:
+    if _use_proxy_iam():
+        user_env = _cloudsql_user_env()
+        _require_proxy_env(user_env)
+        return _proxy_database_uri(user_env), {}
+
+    return _local_database_uri(), {}
+
```
jimmypalelil: The changes you made in [this PR](https://github.com/bcgov/STRR/pull/1764) should work here as well right? we could remove all these lines and simplify

Jacky-Pham: yep this changes the earlier proxy direction. i'll use the shared python connector and pg8000 helper, add migration username support to #77, then repin and relock.

strr-api/src/strr_api/config.py
Jacky-Pham: we ended up using the shared connector here so this now matches #1764

strr-api/src/strr_api/config.py
Jacky-Pham: done this now uses the shared connector for both the api and migrations and all 530 tests and ci pass

PR #1764, opened 2026-07-08
Move remaining STRR services and jobs to Cloud SQL IAM auth

Jacky-Pham:
## What changed

I moved the remaining STRR DB clients over to Cloud SQL automatic IAM auth:

- `strr-email`
- `strr-pay`
- DB-using jobs under `jobs/`
- `interactions-update`, with the env names standardized and the existing `DB_USER==` vault typo fixed

Deployed GCP config now uses the Cloud SQL Python Connector with `enable_iam_auth=True`, `CLOUDSQL_INSTANCE_CONNECTION_NAME`, `DATABASE_NAME`, `DATABASE_USERNAME`, and `CLOUDSQL_IP_TYPE=PUBLIC`.

Local dev remains on the existing password/socket path as long as `CLOUDSQL_INSTANCE_CONNECTION_NAME` is unset and the app is not running on Cloud Run.

## 1Password / rollout notes

I will need the 1Password values updated before these are rolled out:

- Queue services should use `DATABASE_IAM_USERNAME` for `sa-api@bcrbk9-<env>.iam`
- Jobs should use `DATABASE_JOB_IAM_USERNAME` for `sa-job@bcrbk9-<env>.iam`
- `DATABASE_INSTANCE_CONNECTION_NAME` should point at the STRR Cloud SQL instance connection name
- `DATABASE_NAME` should stay `strr-db`
- `CLOUDSQL_IP_TYPE` should stay `PUBLIC`
- UAT should use the `bcrbk9-test` identity/instance because UAT deploys into the test project
- Deployed password/socket DB vars can be removed from these mappings once the 1Password side is ready

## Heads up

The jobs run as `sa-job`. A few of these jobs write to the DB, so if `sa-job` is still readonly on `strr-db`, those jobs will connect with IAM but fail once they try to write. Before rolling out the writing jobs, SRE needs to either give `sa-job` the right DB role or decide which jobs should stay off IAM until that grant is fixed.

## Testing

- `python3 -m compileall` on the touched config/database/test files
- `poetry run pytest tests/unit/test_config_cloudsql_iam.py --no-cov` in `queue_services/strr-email`
- `poetry run pytest tests/unit/test_config_cloudsql_iam.py --no-cov` in `jobs/auto-approval`
- `poetry run pytest tests/unit/test_database_cloud.py tests/unit/test_database.py::test_cloud_sql_missing_vars_raises_error tests/unit/test_gcp_iam_vaults.py --no-cov` in `jobs/interactions-update`
- `git diff --check`


jobs/auto-approval/src/auto_approval/config.py
```diff
@@ -29,6 +29,85 @@
 }
```
jimmypalelil, line 32: a lot of the code has been duplicated here across jobs and services. can we centralize the logic so that such changes are easier to manage

Jacky-Pham: then it makes all the jobs dependent on that one util. if this is okay we can do it but we should note that down that if the helper breaks then all jobs will fail.

jimmypalelil: thats fine -- i think thats better than duplicating

Jacky-Pham: Addressed in 43f72c9c: all affected jobs and queue services now use the shared `bcgov/sbc-connect-common` `cloud-sql-connector` (`DBConfig`/`getconn`) instead of duplicating connector setup and lifecycle logic. The direct connector dependencies were replaced where applicable, each lockfile pins the shared utility revision, and the focused IAM configuration and vault tests pass.

jobs/auto-approval/src/auto_approval/config.py
```diff
@@ -29,6 +30,70 @@
...
+
+    return f"postgresql+pg8000://{db_user}:{db_password}@{db_host}:{db_port}/{db_name}"
+
+
+def _database_settings() -> tuple[str, dict]:
+    if _use_cloudsql_iam():
+        _require_cloudsql_env()
+        return "postgresql+pg8000://", _cloudsql_engine_options()
+
+    return _local_database_uri(), {}
```
jimmypalelil: aren't all these lines simply duplicated in every config.py files in other services. this is the configuraiton i was pointing to before when i asked for centralization. sorry if that wasnt clear enough. lemme know if you have any questions

Jacky-Pham: Addressed in 7cf048e6. I moved the full local/Cloud SQL IAM environment-to-SQLAlchemy configuration into the shared `cloud-sql-connector` package (bcgov/sbc-connect-common#77), replaced the duplicated blocks in all seven jobs and both queue services, and pinned the helper commit in each lockfile. Shared helper tests pass (19/19); auto-approval passes 21/21; the focused strr-email config tests pass 3/3.

Jacky-Pham:
CI container builds could not fetch the companion commit from the upstream repository before bcgov/sbc-connect-common#77 merges, so the dependency is now pinned to the exact commit on my fork. Once the companion PR merges, I will switch the source URL back to bcgov before this PR is merged.

jobs/auto-approval/tests/unit/test_config_cloudsql_iam.py
```diff
@@ -0,0 +1,88 @@
+"""Tests for auto-approval Cloud SQL IAM database configuration."""
+
+import importlib
+
+import pytest
```
jimmypalelil: do we need this test file now that the code is moved out of this repo. we can just add a common fixture and reuse that

Jacky-Pham: agreed i'll remove the duplicate tests, keep the dotenv safe STRR wiring coverage, and move the migration case into #77.

jobs/auto-approval/src/auto_approval/config.py
Jacky-Pham: this is fully moved into #77 now and the app configs only keep their local wiring

jobs/auto-approval/tests/unit/test_config_cloudsql_iam.py
Jacky-Pham: done i kept the small wiring tests here and moved the shared behavior into #77 and all 256 local tests pass and ci is green

shared/python/cloud-sql-connector/src/cloud_sql_connector/connector.py
```diff
@@ -0,0 +1,253 @@
...
+    values = env if env is not None else os.environ
+    db_user = quote(values.get(username_env, ""), safe="")
+    db_password = quote(values.get(password_env, ""), safe="")
+    db_name = values.get(name_env, "")
+    db_host = values.get(host_env, "")
+    db_port = values.get(port_env, "5432")
+
+    if db_unix_socket := values.get("DATABASE_UNIX_SOCKET"):
+        return (
+            f"postgresql+pg8000://{db_user}:{db_password}@/{db_name}"
```
jimmypalelil: why are we swithcing from psycopg2 to pg8000

Jacky-Pham: These jobs already use pg8000 on main. This PR keeps that driver and moves the connection setup into the shared IAM helper. The Google connector version we're using supports pg8000 and asyncpg, but not psycopg2. The existing psycopg2 paths for local tests are unchanged.

jobs/registration_expiry/pyproject.toml
```diff
@@ -18,6 +18,7 @@ sentry-sdk = "^1.44.1"
 flask-babel = "^4.0.0"
 gunicorn = "^21.2.0"
 pg8000 = "^1.31.2"
+cloud-sql-connector = { path = "../../shared/python/cloud-sql-connector" }
```
jimmypalelil: is this path going to work during CD? the other paths are using github link right so do you think we should align with that?

jobs/batch-permit-validator/src/batch_permit_validator/config.py
```diff
@@ -148,10 +137,13 @@ class TestConfig(_Config):  # pylint: disable=too-few-public-methods
     DATABASE_TEST_HOST = os.getenv("DATABASE_TEST_HOST", "")
     DATABASE_TEST_PORT = int(os.getenv("DATABASE_TEST_PORT", "5432"))
 
-    SQLALCHEMY_DATABASE_URI = (
-        f"postgresql://{DATABASE_TEST_USERNAME}:{DATABASE_TEST_PASSWORD}"
-        f"@{DATABASE_TEST_HOST}:{DATABASE_TEST_PORT}/{DATABASE_TEST_NAME}"
-    )
+    # Cloud Run keeps IAM even when a test configuration is selected.
+    if "creator" not in _Config.SQLALCHEMY_ENGINE_OPTIONS:
```
jimmypalelil: is this a fallback method for local testing? can this be refactored into the shared util file instead? -- i see this logic being duplicated across multiple config.py files

scripts/prepare-python-build-context.py
```diff
@@ -0,0 +1,50 @@
+"""Stage one application and its STRR shared package for Docker or Cloud Build."""
```
jimmypalelil: can we not use this build context if uisng the github path for the shared util dependency

PR #1765, opened 2026-07-10
feat(interactions-job-update): modify to job to get email delivery statuses on each notification reference and save it back to database

jimmypalelil:
*Issue:*

- bcgov/strr/issues/

*Description of changes:*


jobs/interactions-update/src/interactions_update/handlers/interaction_status.py
```diff
@@ -0,0 +1,126 @@
...
+    FAILURE_STATUSES = {
+        "FAILURE",
+        "FAILED",
+        "PERMANENT_FAILURE",
+        "TEMPORARY_FAILURE",
+        "TECHNICAL_FAILURE",
+        "VIRUS_SCAN_FAILED",
+    }
+
+    DELIVERED_STATUSES = {"DELIVERED", "SENT"}
```
Jacky-Pham: Could we avoid treating `SENT` as delivered here? It looks like `SENT` only means the provider accepted the email. The final delivery status may come later, so we could miss a subsequent failure.

jobs/interactions-update/src/interactions_update/services/interaction_processor.py
```diff
@@ -0,0 +1,454 @@
...
+                )
+
+                if notify_reference == interaction.notify_reference:
+                    primary_payload = data
+
+            if not recipient_statuses:
+                return None
+
+            existing_meta_data["notify_references"] = ",".join(notify_references)
+            existing_meta_data["notify_delivery"] = {
```
Jacky-Pham, line 296: Could we preserve the recipient statuses from previous runs here? If one lookup temporarily fails, replacing the existing statuses may remove a result we already saved and leave the interaction stuck in `SENT`.

jobs/interactions-update/src/interactions_update/clients/notify_api.py
```diff
@@ -0,0 +1,103 @@
...
+
+        def _make_request():
+            logger.debug(
+                "strr.interactions.notify_request interaction_id=%s notify_reference=%s url=%s",
+                interaction_id,
+                notify_reference,
+                url,
+            )
+            t0 = time.monotonic()
+            resp = requests.get(url, headers=headers, timeout=self.timeout)
```
Jacky-Pham, line 68: Could we apply the configured rate limit before making this request? `NOTIFY_API_RATE_LIMIT_PER_SECOND` is currently loaded and logged, but it does not appear to be enforced, so the worker pool may send more requests than expected.

jobs/interactions-update/src/interactions_update/handlers/interaction_status.py
jimmypalelil: This sent status is actually from GC Notify where it indicates that the upstream server has successfully processed the request. its different from our internal status. But I have removed it from the list to avoid confusion.

ref: https://documentation.notification.canada.ca/en/status.html#email-status

jobs/interactions-update/src/interactions_update/services/interaction_processor.py
jimmypalelil: if its sent then it would get picked up again and re-processed and the failures would be reconciled

jobs/interactions-update/src/interactions_update/clients/notify_api.py
jimmypalelil: retry policy is applied on the notify api client and thats whats used in every call: https://github.com/jimmypalelil/STRR/blob/73a597a76de04209bb6a48bad405e9b1fd4e15f9/jobs/interactions-update/src/interactions_update/clients/notify_api.py#L90-L101

Jacky-Pham (approved):
LGTM

PR #1768, opened 2026-07-14
fix(batch-permit-job): use runtime auth for storage

Jacky-Pham:
## Summary
- remove the batch permit validator job's deployed GCP_AUTH_KEY mapping so it uses the Cloud Run runtime identity
- support keyless signed URL generation by passing refreshed ADC access token and service account email to GCS
- add focused unit coverage for the keyless signed URL path


strr-api/src/strr_api/services/gcp_storage_service.py
jimmypalelil, line 65: the auth key is being used here but removed above. this will result in job failures

strr-api/src/strr_api/services/gcp_storage_service.py
```diff
@@ -190,11 +191,27 @@ def upload_file(cls, file_type, file_contents, bucket_id, metadata=None):
     @classmethod
     def get_presigned_url(cls, bucket_id, blob_name, expiration_minutes):
         """Gets the presigned url for a file."""
-        bucket = cls.get_bucket(bucket_id)
+        storage_client = cls._create_storage_client()
+        bucket = storage_client.bucket(bucket_id)
```
jimmypalelil: get_bucket method already exists above

strr-api/src/strr_api/services/gcp_storage_service.py
```diff
@@ -190,11 +191,27 @@ def upload_file(cls, file_type, file_contents, bucket_id, metadata=None):
...
         blob = bucket.blob(blob_name)
+        signed_url_kwargs = {}
+
+        credentials = storage_client._credentials  # pylint: disable=protected-access
+        if credentials and not isinstance(credentials, service_account.Credentials):
+            credentials.refresh(Request())
```
jimmypalelil: do we really need to refresh the credentials. wont google manage that automatically since the job is running on **GCP Cloud Run**?

strr-api/src/strr_api/services/gcp_storage_service.py
Jacky-Pham: yeah good catch, updated this to use get_bucket

strr-api/src/strr_api/services/gcp_storage_service.py
Jacky-Pham: yeah Google handles the normal storage requests, but this signed URL path passes the token directly into the IAM signBlob call. I updated it so we only refresh when the ADC token isn’t valid

strr-api/src/strr_api/services/gcp_storage_service.py
Jacky-Pham: yeah, when the key isn’t set this falls back to the Cloud Run service account through ADC. I also added the IAM change it needs for signed URLs in bcgov/bcregistry-sre#393

PR #1769, opened 2026-07-14
feat(strr-api): include email interactions in events payload

jimmypalelil:
*Issue:*

JIRA: https://hous-hpb.atlassian.net/browse/REGBACKLOG-74

*Description of changes:*
In the existing events endpoint, email interaction events are added so that the filing history time line view can support these event rows on the UI

---

### DIFF
--- a/strr-api/pyproject.toml
+++ b/strr-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "strr-api"
-version = "0.3.29"
+version = "0.3.30"
 description = ""
 authors = ["thorwolpert <thor@wolpert.ca>"]
 license = "BSD 3-Clause"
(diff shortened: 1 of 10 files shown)


Jacky-Pham (commented):
Restrict email delivery details to staff

Public users can turn on this option and see email addresses, delivery failures, and provider references. These details i assume are intended for staff. Only allow staff/system users to request them, or remove sensitive fields from the public response.

application.py the code adds interaction details without checking whether the user is staff.

The same issue also exists at:
registrations.py around line 477

jimmypalelil:
> Restrict email delivery details to staff
> 
> Public users can turn on this option and see email addresses, delivery failures, and provider references. These details i assume are intended for staff. Only allow staff/system users to request them, or remove sensitive fields from the public response.
> 
> application.py the code adds interaction details without checking whether the user is staff.
> 
> The same issue also exists at: registrations.py around line 477

non - staff or system users will only be able to view the applications and registration events and interactions that they have created or they have access to. In this case it should be fine if they are able to see the email addresses and other details.

Jacky-Pham (approved):
as per Jimmy's earlier comment it makes sense to me. Approved

PR #1773, opened 2026-07-27
fix(examiner-ui): modify the e2e tests so they all pass

jimmypalelil:
_Issue:_

- JIRA: https://hous-hpb.atlassian.net/browse/REGBACKLOG-72

_Description of changes:_
- Modified the e2e tests so they all pass
- bumped up the version number in examiner ui
- tested the full e2e auth setup with bceid for examiner-ui and verified that they all pass given the correct credentials

---


Jacky-Pham:
Can we make this test select a host application instead of always selecting the first application? The first application could be a platform or strata application, which would cause the host-only assertions to fail depending on the environment’s data.

Also, the sync repositories changes in both .vscode/tasks.json files don’t appear related to this PR. Can we remove them?

jimmypalelil:
The work to refine the tests need to be done later.. i am thinking of seeding the db with python scripts just for e2e and testing against that so that its deterministic.

The sync repo task is a useful repo task to have in the codebase. i use it all the time. Didn't want to create another ticket just for adding this new task entry. any other reason why it should be removed?

Jacky-Pham (approved):
Nope seems we have a plan in place thats all that matters. 
Approved :)

PR #1774, opened 2026-07-27
refactor(strr-api): update get_registration_document_by_key in document_service to use first() and add tests

jimmypalelil:
_Issue:_

- JIRA: https://hous-hpb.atlassian.net/browse/REGBACKLOG-341

_Description of changes:_

### DocumentService Updates & Tests

- **Refactor:** Updated `get_registration_document_by_key` in `DocumentService` to use `.first()` instead of strict single-record lookups to gracefully handle cases where multiple documents share the same path key.
- **Unit Tests:** Added a `pytest` test suite for `DocumentService` using `unittest.mock` to isolate database and GCP interactions:
  - Added unit test coverage for uploading, deleting, and fetching registration documents.
  - Added an edge case test ensuring `get_registration_document_by_key` correctly retrieves the first document when duplicate path keys exist in the database.

---

### DIFF
--- a/strr-api/src/strr_api/services/document_service.py
+++ b/strr-api/src/strr_api/services/document_service.py
@@ -78,9 +78,7 @@ class DocumentService:
     def get_registration_document_by_key(cls, registration_id, file_key):
         """Get registration document by key."""
         return (
-            Document.query.filter(Document.registration_id == registration_id)
-            .filter(Document.path == file_key)
-            .one_or_none()
+            Document.query.filter(Document.registration_id == registration_id).filter(Document.path == file_key).first()
         )
 
     @classmethod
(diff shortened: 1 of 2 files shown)


Jacky-Pham:
Can we add an order_by before calling .first()? If multiple documents have the same file key, the database could return any of them. Since the selected document’s filename and file type are used for the download, we should decide whether to use the oldest or newest document and make that explicit.

jimmypalelil:
> Can we add an order_by before calling .first()? If multiple documents have the same file key, the database could return any of them. Since the selected document’s filename and file type are used for the download, we should decide whether to use the oldest or newest document and make that explicit.

Its not required to sort them and get the latest. if there are multiple rows for the same file id, they are supposed to be the exact same file in the GCS bucket. its just that the entries in the db might be duplicates. We are planning to address the root cause of this duplicate creation but in the meantime its better to serve the file

Jacky-Pham (approved):
LGTM

